PR #5, opened 2026-04-20
Introduce ContainerRuntime abstraction

andrewlaack:
Phase 0 of the rootless Podman migration. Moves every docker CLI call behind a ContainerRuntime protocol so future runtimes (rootless Podman, gVisor-backed runc, etc.) can slot in without touching callers.

What changed

- New compute_space/core/runtimes/ package with a base Protocol, the extracted DockerRuntime implementation, and a factory keyed off a new container_runtime config option. Default is docker, so behavior on existing deployments is unchanged.
- containers.py keeps its public function surface; each free function delegates to get_runtime().
- Renames the build-cache-corrupt error marker to [BUILD_CACHE_CORRUPT] and the API error_kind to build_cache_corrupt. The UI accepts both the new and legacy values so an upgraded router talking to a cached browser still works.
- Unit tests for the factory, the Docker runtime's subprocess invocations, and the delegation layer.

What is not in this PR

- No DB migration. The docker_container_id column stays untouched. It will be renamed in Phase 2 alongside adding uid_map_base for per-app uidmap allocation.
- No build-log rename. docker.log stays as the filename for this phase.
- No ansible changes.
- No Podman code. This is only the abstraction scaffolding.

Verification

- uv run pytest : 200 passed, 30 skipped (docker/tls-gated).
- uv run ruff check . : clean.
- uv run ruff format --check . : clean.
- uv run mypy : clean.
- vet : clean.

### DIFF
--- a/compute_space/compute_space/config.py
+++ b/compute_space/compute_space/config.py
@@ -43,6 +43,10 @@ class Config:
     port_range_start: int
     port_range_end: int
 
+    # Container runtime: which runtime manages app containers.
+    # Supported values are defined in compute_space.core.runtimes.factory.
+    container_runtime: str
+
     def evolve(self, **kwargs: Any) -> Self:
         return attr.evolve(self, **kwargs)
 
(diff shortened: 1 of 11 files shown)


andrewlaack:
Superseded: switching to a straight Docker->Podman replacement rather than a runtime abstraction. New PR to come.

PR #6, opened 2026-04-20
Ansible task for rootless podman bootstrap

andrewlaack:
Phase 1 of the rootless Podman migration.  Stacked on #5.

Not wired into setup.yml by default. Opt in per-server with -e openhost_container_runtime=podman.

What this adds

- tasks/podman.yml installs podman + crun + pasta + slirp4netns + netavark + aardvark-dns + uidmap + fuse-overlayfs, allocates a 10M subuid/subgid range to the host user, enables user lingering (so restart=unless-stopped containers survive reboot), lowers net.ipv4.ip_unprivileged_port_start to 80, and fails the play if the data filesystem does not support idmapped mounts.
- setup.yml selects docker.yml or podman.yml based on the runtime variable.  Default remains docker, so existing bootstrap is unchanged.
- config.toml.j2 emits container_runtime so the router's runtime factory picks the right backend.

What is not in this PR

- No PodmanRuntime implementation. That is Phase 2. Running the ansible with -e openhost_container_runtime=podman today will boot the host correctly and then the router will fail with a clear 'Unknown container_runtime' error at startup. That is intentional guardrail behavior.
- No vm-manager template change. The default image still provisions with Docker.

Verification

- ansible-playbook --syntax-check passes for both runtime selections.
- Rendered config.toml round-trips through typed_settings.load for both runtimes.
- uv run pytest: 200 passed, 30 skipped.  Ruff and mypy clean.
- vet clean.
- Not yet runtime-tested on a live VM (pending vm-manager integration test in Phase 4).

### DIFF
--- a/ansible/readme.md
+++ b/ansible/readme.md
@@ -16,6 +16,14 @@ ansible-playbook ansible/setup.yml -i <IP>, -e domain=<domain> -e initial_user=r
 
 the trailing comma after the IP is required (tells ansible it's a host list, not a file).
 
+### container runtime
+
+by default the bootstrap installs docker ce and adds the host user to the docker group. pass `-e openhost_container_runtime=podman` to install rootless podman instead. this is the recommended path for new servers going forward; docker remains the default for now until the migration is fully validated. the two are mutually exclusive per server.
+
+```bash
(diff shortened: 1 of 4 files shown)


andrewlaack:
Superseded: switching to a straight Docker->Podman replacement rather than a runtime abstraction. New PR to come.

PR #8, opened 2026-04-20
Replace docker with more secure podman

andrewlaack:
Replaces Docker CE with rootless Podman as the container runtime for every app.

Runtime

Ansible installs podman, crun, passt, slirp4netns, netavark, aardvark-dns, uidmap and fuse-overlayfs for the host user; enables systemd user lingering so containers survive reboots; lowers net.ipv4.ip_unprivileged_port_start to 80 so apps can publish 80/443 without root; asserts the data filesystem supports idmapped bind mounts.  Apps run in podman's default single-user namespace.  Every bind mount uses :idmap (or :ro,idmap) so container-root writes land on disk owned by the host user without chmod 0o777.  Containers start with --cap-drop=ALL, --security-opt=no-new-privileges=true, and both --add-host=host.docker.internal:host-gateway and host.containers.internal so existing apps keep working.

Manifest validation

SAFE_CAPABILITIES and SAFE_DEVICE_PATHS allowlists reject unsafe caps (SYS_ADMIN, SYS_MODULE, SYS_PTRACE, …) and unsafe devices (/dev/mem, /dev/kvm, raw block devices, …) at parse time.  host_port values below 80 are rejected at parse time.  Capability names accept both NET_ADMIN and CAP_NET_ADMIN forms.

Upgrade safety for existing instances

Existing Docker-based instances that click the dashboard Update button end up running the new code against a host that has no podman.  To avoid a crashed router and a dark dashboard:

- compute_space.core.containers.podman_available() probes podman --version without raising.
- core.startup._check_app_status refuses to rebuild when podman is missing and instead marks every running/starting/building app as status='error' with a clear PODMAN_MISSING_ERROR remediation message.  The dashboard stays reachable so the operator can see what to do.
- ansible/tasks/podman.yml writes /etc/openhost/runtime with a runtime_version integer; core.runtime_sentinel reads it so future host-side changes (sysctl, sudoers, allowlist) can signal required re-prep even when the podman binary probe alone wouldn't notice.
- /api/settings/check_for_updates returns a host_prep_ok field; the settings UI shows a banner and disables the Update button when host prep is needed.  /api/settings/update_repo_state refuses with HTTP 409 under the same condition so direct curl / stale pages can't bypass the UI guard.

Data handling

Shared rmtree_with_sudo_fallback helper replaces the old docker-run-alpine-rm-rf cleanup (opt-in raise_on_failure for code-sync paths, log-and-swallow for deprovision paths).  provision_data no longer chmods app data dirs to 0o777.

Database

docker_container_id → container_id.  The migration renames before any _recreate_table step so data isn't silently dropped by the common-column filter; covered by regression tests.

Tests and CI

New suites: test_containers.py, test_podman_availability.py, test_settings_host_prep.py, test_data_rmtree.py, test_podman_migration.py, test_runtime_sentinel.py, test_app_status.py, plus capability / device / port-floor coverage in test_manifest.py.  requires_docker / --run-docker renamed to requires_podman / --run-podman.  New test-podman CI job installs rootless podman on ubuntu-latest and runs the podman-aware unit suites.  The idmap-dependent @requires_podman e2e classes still run on real VMs via the e2e workflow.

Docs

README, claude.md, creating_an_app.md and manifest_spec.md updated.  Rootless-specific constraints (port floor, cap / device allowlists) are documented so app authors see them without reading the code.


compute_space/compute_space/web/start.py
```diff
@@ -34,10 +43,45 @@ def _terminate_children(children: list[subprocess.Popen[bytes]]) -> None:
             proc.kill()
 
 
+def _enforce_runtime_sentinel_if_requested() -> None:
+    """Gate startup on /etc/openhost/runtime when running under systemd.
+
+    The ansible playbook sets ``OPENHOST_CHECK_RUNTIME_SENTINEL=1`` in the
```
andrewlaack: I dislike this approach to non-backwards compatible updates, but I'm not sure what else to do.

ansible/setup.yml
```diff
@@ -27,7 +27,7 @@
     ansible_python_interpreter: /usr/bin/python3
   tasks:
     - import_tasks: tasks/apt_packages.yml
-    - import_tasks: tasks/docker.yml
+    - import_tasks: tasks/podman.yml
```
andrewlaack, line 30: Ideally, users will re-run this to update their spaces. alternatively they could run some basic podman install things which should work (TM).

andrewlaack:
Manual testing of these changes:

1. set your update branch to this branch
2. update the code
3. run ansible-playbook
  a. installs podman + friends
5. restart vps
  a. you could probably just kill all docker containers and then restart the compute space, but this is more commands, and this is where I've noticed openhost systemd service not starting correctly on the first systemctl start openhost. 

For users, we would ssh in, run playbook, reboot. This should just work TM

ansible/tasks/podman.yml
```diff
@@ -0,0 +1,122 @@
...
+  stat:
+    path: /var/lib/systemd/linger/host
+  register: _linger_sentinel
+
+- name: Enable user lingering for host
+  command: loginctl enable-linger host
+  when: not _linger_sentinel.stat.exists
+  changed_when: true
+
+# Rootless podman needs to publish ports >= 80 (the router's [[ports]]
```
zplizzi: i think we need to set this to 25 so we can get SMTP

compute_space/compute_space/core/containers.py
```diff
@@ -1,3 +1,22 @@
+"""Container lifecycle using rootless Podman.
+
+Every container runs under the unprivileged ``host`` user in podman's
+default rootless user namespace.  Host bind mounts use ``:idmap`` so
```
zplizzi, line 4: does this auto-detect the user being used in the docker container?

compute_space/compute_space/core/containers.py
```diff
@@ -191,64 +304,84 @@ def run_container(
...
+    logger.info("Dropping container build cache: %s", " ".join(cmd))
     result = subprocess.run(cmd, capture_output=True, text=True, timeout=120)
     output = (result.stdout + result.stderr).strip()
     if result.returncode != 0:
-        raise RuntimeError(output or "Docker builder prune failed")
+        raise RuntimeError(output or "podman image prune failed")
     return output
 
 
 def get_container_status(container_id: str) -> str:
```
zplizzi: can we convert this into some proper type? enum or dataclass or just a bool if all we care about is running or not.

compute_space/compute_space/core/data.py
```diff
@@ -3,12 +3,66 @@
...
 from compute_space.core.logging import logger
 from compute_space.core.manifest import AppManifest
 
 
+def rmtree_with_sudo_fallback(path: str, *, raise_on_failure: bool = False) -> None:
+    """Remove a directory tree, falling back to ``sudo -n rm -rf``.
+
+    The fast path is ``shutil.rmtree`` with an onexc hook that chmods
+    read-only entries (git-clone artefacts) and retries.  The sudo
+    fallback relies on the NOPASSWD sudoers rule installed by ansible.
```
zplizzi, line 18: fwiw i hope to remove this eventually

compute_space/compute_space/core/data.py
```diff
@@ -74,46 +128,13 @@ def provision_data(
 
 
 def _remove_dir(dir_path: str) -> None:
```
zplizzi: do we need this func now?

compute_space/compute_space/core/manifest.py
```diff
@@ -8,6 +8,63 @@
...
+    }
+)
+
+# Host devices safe to pass through to rootless containers.  Anything
+# outside this set (e.g. /dev/mem, /dev/kmem, raw block devices, /dev/kvm)
+# is rejected at parse time.
+SAFE_DEVICE_PATHS: frozenset[str] = frozenset(
+    {
+        "/dev/net/tun",
+        "/dev/random",
```
zplizzi: any reason not to expose some of these by default? are these not available in docker? eg /dev/random and /dev/null are pretty commonly used?

compute_space/compute_space/core/runtime_sentinel.py
```diff
@@ -0,0 +1,128 @@
...
+SENTINEL_PATH: Final[str] = "/etc/openhost/runtime"
+
+# Bump ``EXPECTED_RUNTIME_VERSION`` (here) and the value written by
+# ``ansible/tasks/podman.yml`` together whenever a host-side change
+# ships that existing hosts must adopt before their next router start.
+EXPECTED_RUNTIME: Final[str] = "podman"
+EXPECTED_RUNTIME_VERSION: Final[int] = 1
+
+
+@dataclass(frozen=True)
```
zplizzi: i'm trying to migrate us to attr.s(auto_attrs=True, frozen=True)

compute_space/compute_space/core/startup.py
```diff
@@ -7,32 +7,66 @@
 from compute_space.config import Config
 from compute_space.core import identity
 from compute_space.core.apps import start_app_process
+from compute_space.core.containers import PODMAN_MISSING_ERROR
```
zplizzi: can we make these like `CONTAINER_RUNTIME_MISSING_ERROR` just to keep more generic in case we change again.

compute_space/compute_space/core/startup.py
```diff
@@ -7,32 +7,66 @@
...
 from compute_space.core.containers import get_container_status
+from compute_space.core.containers import podman_available
```
zplizzi: same here

compute_space/compute_space/core/startup.py
```diff
@@ -7,32 +7,66 @@
...
 from compute_space.core.logging import logger
 from compute_space.core.storage import start_storage_guard
 from compute_space.db import init_db
 
 
+def _mark_running_apps_podman_missing(config: Config) -> int:
```
zplizzi: and here

zplizzi: and elsewhere in this file

compute_space/compute_space/web/routes/api/settings.py
```diff
@@ -71,6 +74,27 @@ async def set_remote() -> ResponseReturnValue:
     return jsonify({"ok": True, "token_applied": token_applied})
 
 
+def _host_prep_payload() -> dict[str, object]:
+    """Return whether the host is ready to run the installed router code.
+
+    Combines a live ``podman --version`` probe with the
+    ``/etc/openhost/runtime`` sentinel.  Never raises.
+    """
+    podman_ok = podman_available()
```
zplizzi: i think this should be made generic with `container_runtime` vs `podman`. ideally only the container-related files have to know anything about what runtime we're actually using.

compute_space/compute_space/web/templates/settings.html
```diff
@@ -51,22 +51,38 @@ <h3>Restart</h3>
 
     const checkAgainBtn = '<button onclick="checkForUpdates()" class="btn" style="margin-top:0.5em;">Check again</button>';
 
+    // Banner when host prep is needed; Update button is disabled below.
+    let hostPrepBanner = '';
+    if (data.host_prep_ok === false) {
+      hostPrepBanner = '<div class="error" style="margin-bottom:0.75em;padding:0.5em;border:1px solid #c00;">'
+        + '<strong>Host not prepared for this router version.</strong> '
```
zplizzi, line 58: this is nice, thanks

compute_space/compute_space/web/proxy.py
```diff
@@ -176,15 +177,23 @@ async def ws_proxy(
     # the handshake completes and both send/receive are immediately usable.
     await client_ws.accept()
 
+    # Only pass `subprotocols` if the client actually negotiated some.
```
zplizzi: nit - seems like your auto-comment-wrapper is set to a narrower width than our code formatter?

compute_space/tests/conftest.py
```diff
@@ -88,15 +101,15 @@ def _stop_router_process(proc: subprocess.Popen[Any]) -> None:
         proc.wait()
 
 
-def _docker_cleanup(container_name: str, app_name: str) -> None:
-    """Force-remove a test Docker container and image."""
+def _podman_cleanup(container_name: str, app_name: str) -> None:
```
zplizzi: can this move into container.py maybe? to keep all runtime-specific code together

zplizzi: and call container_cleanup or something

zplizzi: and maybe take a quick pass thru the rest of the tests and try to minimize references to `podman` specifically, replacing with generic references to `container_runtime` or something and making helpers as necessary in container.py to expose generic interfaces.

zplizzi: doesn't need to be perfect, just directionally moving in that direction.

docs/creating_an_app.md
```diff
@@ -6,15 +6,23 @@ This guide walks through building an app that runs on OpenHost.
...
-Apps can be anything that can run in a docker container, and accessed via HTTP(s).
+Apps can be anything that can run in an OCI container, and accessed via HTTP(s). OpenHost runs every app under rootless podman, so container-root maps to an unprivileged subuid on the host rather than real root.
 
 An `openhost.toml` manifest must be placed at the root of your repo, to indicate to OpenHost how to run your app. See the [manifest spec](manifest_spec.md) for the full field reference.
 
+### Rootless constraints
+
+A few things that work under classical Docker don't work here:
+
+- `[[ports]].host_port` values below 80 are rejected at manifest parse time — rootless podman cannot bind to privileged ports under 80 (the router lowers the unprivileged-port floor from 1024 to 80 so HTTP/HTTPS still work).
```
zplizzi: update here too

docs/creating_an_app.md
```diff
@@ -99,7 +107,7 @@ The router injects these environment variables into your app:
 |----------|---------|-----------------------------------------------------------------------------------------------------------------|
 | `OPENHOST_APP_NAME` | `my-app` | Your app's name, as registered with OpenHost. This will be the subdomain the app is routeable at.               |
 | `OPENHOST_APP_TOKEN` | `kF3xP_2qA-bN4...` (43-char url-safe token) | Random per-app token used to authenticate cross-app service calls                                               |
-| `OPENHOST_ROUTER_URL` | `http://host.docker.internal:8080` | internal URL of the router, used for constructing service requests                                              |
+| `OPENHOST_ROUTER_URL` | `http://host.docker.internal:8080` | internal URL of the router, used for constructing service requests. Resolves via `host.docker.internal` (kept for compatibility) or equivalently `host.containers.internal`. |
```
zplizzi: i'm not sure we need the new part here? apps shouldn't care what this URL is set to, as long as it works.

docs/manifest_spec.md
```diff
@@ -20,14 +20,16 @@ Apps declare how they should be deployed on OpenHost by placing an `openhost.tom
...
+| `devices` | string[] | no | `[]` | Host devices to pass through (e.g., `"/dev/net/tun"`). Restricted to a rootless-safe allowlist (see `compute_space.core.manifest.SAFE_DEVICE_PATHS`); paths like `/dev/mem`, `/dev/kvm`, or raw block devices are rejected at parse time. |
 
 ### `[[ports]]` — optional, repeatable
 
 Declares additional port mappings for the container. Each entry binds a container port to a host port (TCP+UDP on 0.0.0.0). Set `host_port = 0` for auto-assignment from the 9000-9999 range.
 
+Rootless podman can bind ports >= 80 only; `host_port` values below 80 are rejected at parse time. Use `host_port = 80` / `443` / any value from the auto-assign range, or route through the built-in router proxy.
```
zplizzi: and update here

zplizzi: also aren't `80` / `443` reserved for the router?

docs/manifest_spec.md
```diff
@@ -79,7 +81,7 @@ The host provisions requested data services and injects connection info as envir
 - `OPENHOST_APP_DATA_DIR` — `/data/app_data/{app_name}` (only if app_data access granted)
 - `OPENHOST_APP_TEMP_DIR` — `/data/app_temp_data/{app_name}` (only if app_temp_data access granted)
 - `OPENHOST_AUTH_PUBLIC_KEY` — PEM-encoded JWT public key for token verification (only if signing keys are available)
-- `OPENHOST_ROUTER_URL` — URL of the router's HTTP server (e.g., `http://host.docker.internal:<port>`)
+- `OPENHOST_ROUTER_URL` — URL of the router's HTTP server. Points at `host.docker.internal` (kept for compatibility) or equivalently `host.containers.internal`; both resolve to the host gateway from inside the container.
```
zplizzi: same comment here

self_host_cli/doctor.py
```diff
@@ -41,20 +42,40 @@ def _check_uv() -> _Check:
     return _Check("uv installed", False, "uv not found on PATH")
 
 
-def _check_docker() -> _Check:
+def _check_podman() -> _Check:
```
zplizzi: same comments here about making generic to runtime.

claude.md
```diff
@@ -21,15 +21,15 @@ openhost/
...
+1. **compute_space** is a quart app (port 8080). it reads `openhost.toml` manifests from app repos, builds images from each app's `Dockerfile` using rootless podman, and runs each app in its own user namespace.
 2. it proxies incoming HTTP requests to the correct app by matching subdomain or URL path prefix (`/{app_name}`).
 3. **auth** uses JWT with RS256. apps verify with the public key passed as env var.
 
 ## running and testing
 
 - **all lightweight tests**: `uv run --group dev pytest` (from project root)
-- **+ Docker tests**: `uv run --group dev pytest --run-docker`
-- **everything**: `uv run --group dev pytest --run-docker`
+- **+ podman integration tests**: `uv run --group dev pytest --run-podman`
```
zplizzi: let's make this flag `--requires-containers` or something

.github/workflows/ci.yml
```diff
@@ -25,9 +25,34 @@ jobs:
       - uses: astral-sh/setup-uv@v6
       - run: uv run pytest
 
-  test-docker:
+  # Rootless podman unit suites.  The full-stack @requires_podman
```
zplizzi: i don't really understand this or why we need the explicit list of files below. what's the specific issue with github runners? unsupported filesystem?

zplizzi (approved):
lots of stylistic nits but overall this looks good! i didn't realize this would be so involved, appreciate you working thru it. 

i would like to know if we can manage more of this with pixi vs system package manager.

compute_space/compute_space/core/containers.py
andrewlaack: yeah, rootless podman auto-idmaps the container's uid 0 to the host `host` user via the user namespace, and bind mounts use `:idmap` so writes land owned by `host` on disk regardless of what `USER` the image declares.

PR #9, opened 2026-04-20
Unify storage metrics into single disk metric

andrewlaack:
## Summary

- Replace separate persistent/temporary storage metrics with a single unified disk metric since both live on the same filesystem under `data_root_dir`
- Dashboard now shows "Disk free: X / Y" instead of separate persistent/temporary lines
- Rename `persistent_free_bytes` to `disk_free_bytes` throughout
- Update all tests, docs, and error messages to match

Migrated from imbue-ai/openhost#163.


andrewlaack:
we want this since everything is one the same partition now.

PR #13, opened 2026-04-21
Hash refresh tokens and app tokens at rest

kilo-gi:
## Summary

- Store SHA-256 hashes instead of plaintext bearer tokens in `refresh_tokens` and `app_tokens` tables, matching existing `api_tokens` pattern
- Migration detects old `token` column, renames to `token_hash`, and hashes existing plaintext values in place
- All write paths (setup, login, app deploy/start) hash before INSERT; all read paths (refresh, logout, service auth) hash before query

## Test plan

- [x] Migration correctly hashes old plaintext refresh_tokens rows
- [x] Migration correctly hashes old plaintext app_tokens rows
- [x] Migration is idempotent (double init_db doesn't double-hash)
- [x] Stored token_hash != raw token value (SHA-256 hex digest)
- [x] Lookup by hashed value finds correct row
- [x] Raw plaintext token does not match any row
- [x] All existing tests pass (119 passed)

### DIFF
--- a/compute_space/compute_space/core/apps.py
+++ b/compute_space/compute_space/core/apps.py
@@ -5,6 +5,7 @@ Extracted from routes/apps.py — no HTTP/Quart dependencies.
 
 import asyncio
 import dataclasses
+import hashlib
 import json
 import os
 import re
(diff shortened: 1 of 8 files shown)


zplizzi (approved):
lgtm!

PR #17, opened 2026-04-21
Fix login crash when password_hash is NULL

kilo-gi:
## Summary
- Guard against NULL `password_hash` in login route before calling `bcrypt.checkpw`
- Schema allows nullable `password_hash` (for `password_needs_set` flow), but `None.encode()` crashes
- Redirects to setup page when password not yet set

## Test plan
- [ ] Verify login works normally with valid password_hash
- [ ] Verify owner with NULL password_hash redirects to setup instead of crashing


kilo-gi:
Closing, probably superseded by https://github.com/imbue-openhost/openhost/pull/25 since I didn't like the migration code.

PR #23, opened 2026-04-23
Fine-grained filesystem access flags for app manifests

andrewlaack:
Splits the monolithic access_all_data manifest flag into three independent fields so apps can request exactly the filesystem access they need.

New [data] fields:
- access_all_apps_data — RW mount of /data/app_data (every app)
- access_all_apps_temp_data — RW mount of /data/app_temp_data (every app)
- access_vm_data_rw — RW mount of /data/vm_data (access_vm_data stays RO)

Any combination is valid. access_all_data is kept as legacy shorthand for all three of the new flags.

Manifest rejects access_vm_data (RO) combined with either access_vm_data_rw or access_all_data at parse time.

AppManifest gains wants_* properties that centralise the resolution of scoped + broad + legacy flags. containers.compute_data_mounts is extracted as a pure helper returning docker mount tuples; run_container consumes it. provision_data uses the same helpers. 72 unit tests + 7 provision_data cases cover the new behavior.

### DIFF
--- a/compute_space/compute_space/core/containers.py
+++ b/compute_space/compute_space/core/containers.py
@@ -15,6 +15,71 @@ _ANSI_RE = re.compile(r"\x1b\[[0-9;]*[a-zA-Z]|\x1b\([AB0-9]|\x1b[=>]|\x0f|\r")
 _CACHE_CORRUPT_RE = re.compile(r"content digest sha256:[0-9a-f]+: not found")
 
 
+def compute_data_mounts(
+    manifest: AppManifest,
+    app_name: str,
+    data_dir: str,
+    temp_data_dir: str,
+) -> list[tuple[str, str, str | None]]:
+    """Return the list of data-volume mounts implied by a manifest.
+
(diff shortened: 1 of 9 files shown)


andrewlaack:
Follow-up commit (22b4e59) adds `access_openhost_state_ro` to this PR.

Context: while testing full-instance restore via the backup app, restoring a snapshot to a fresh instance left it unable to start any apps. Root cause: `access_all_data = true` only mounts `/data/app_data`, `/data/app_temp_data`, and `/data/vm_data` — the router's own SQLite (`router.db`), TLS material, and signing keys live in `{persistent_data_dir}/openhost/`, which is not reachable by any OpenHost app with the current manifest permissions. So the backup app can't see router state, its snapshots can't include it, and restored instances have app data but an empty apps table.

The new flag adds a read-only mount of `/data/openhost` for apps that opt in explicitly. Kept out of `access_all_data` so existing manifests using that legacy flag do not silently gain access to router tokens and the owner password hash.

andrewlaack:
Closing for now; waiting on services changes to start.

PR #22, opened 2026-04-23
Adopt yoyo-migrations for compute_space DB

kilo-gi:
## Summary

Adopts **yoyo-migrations** for `compute_space`'s SQLite DB, plus a pure-replay snapshot/golden-test harness driven by realistic frozen DB dumps.

**Migration framework**
- SQL-first yoyo migrations under `compute_space/compute_space/db/migrations/`
- DB tracks applied versions via `_yoyo_migration` — no more column-probing guesswork
- `init_db()` dispatches on three DB states: **fresh** (yoyo applies all), **legacy** (`migrate()` then idempotently apply `0001`), **managed** (pending only)
- Existing `migrate()` preserved intact as legacy on-ramp; renamed `migrations.py` → `legacy_migrate.py`
- `schema.sql` retained as human-readable reference; no longer executed at startup
- Added `0002_dummy.py` — noop — so the snapshot chain has a real migration pair to exercise

**Testing**
- Split: `test_legacy_migrations.py` (router's `migrate()` path) and `test_yoyo_migrations.py` (yoyo dispatch + snapshot suite)
- Snapshot harness at `compute_space/tests/fixtures/migrations/` = pure replay. No per-scenario Python builders. Each `at_NNNN.sql` is a self-contained `.dump` carrying schema, app data, and yoyo tracking rows.
- Two realistic scenarios — `scenario_refresh_tokens` (4 token rows × active/revoked × future/past expiry) and `scenario_port_mappings` (2 apps × 3 mappings each)
- Test harness parametrizes over ordered `(at_from, at_to)` pairs strictly between present files — scenarios can cherry-pick checkpoints (`at_0001` + `at_0015` only if desired), only the endpoints are validated
- `regenerate.py` is pure replay: load highest existing snapshot, apply yoyo migrations, dump. `--check` mode for CI drift detection.
- `_EXPECTED_TABLES` constant deleted; `TestYoyoDispatch` compares against fresh-bootstrap schema snapshots
- Deterministic snapshots: fixed timestamps, canonicalised yoyo timestamps, stripped volatile yoyo_log/lock rows

## Process

Full design and review trail in `docs/db_migrations/`:
- `goals.md`, `requirements.md` — scoped + approved up front
- `implementation-notes.md` — what each agent landed
- `review-feedback-round1.md` — flagged DDL duplication (fixed)
- `review-feedback-round2.md` — PASS
- `review-feedback-snapshot.md` — first snapshot harness review, PASS
[...]

PR #25, opened 2026-04-23
Make owner.password_hash NOT NULL and drop password_needs_set

kilo-gi:
Stacked on #22 (feat/yoyo-migrations).

## Summary

- `password_hash` should never be null — `/setup` always writes a bcrypt hash before any request can authenticate. Tighten the schema to reflect that.
- `password_needs_set` was added alongside the nullable column but nothing ever read or cleared it. Drop it.

## Changes

- New Python yoyo migration `0002_drop_password_needs_set.py` recreates `owner` with `password_hash NOT NULL` and no `password_needs_set` column. (SQLite can't alter NOT NULL in place.) Any row stuck in a pre-setup `NULL password_hash` state is deleted — it can't satisfy the new constraint and an owner without a password is unusable.
- `auth.py /setup`: drop `password_needs_set` from the `INSERT`.
- `schema.sql`: updated reference to match post-0002 state.
- Tests: replace the now-obsolete "migration adds `password_needs_set`" test with equivalents that assert the final schema (`password_hash NOT NULL`, no `password_needs_set`, data preserved) and that legacy rows with `NULL password_hash` are dropped.

`login()` already accesses `owner["password_hash"].encode()` without a null-check; post-migration that's provably safe.

## Test plan

- [ ] CI green
- [ ] Manual: legacy DB with `NULL password_hash` row is cleared on startup


kilo-gi:
Superseded by #42 (merged 2026-04-27).

PR #26, opened 2026-04-24
Drop redundant username claim from JWT payload

kilo-gi:
## Summary
- Remove duplicated `username` claim from JWT payload in `create_access_token` and the API-token identity path. Standard `sub` claim already carries the same value.
- Update integration test payload accordingly.

No code reads `username` from claims (verified with grep), so this is safe.

## Test plan
- [ ] `uv run --group dev pytest compute_space/tests/ -v`


zplizzi:
we no longer have JWTs

PR #44, opened 2026-04-28
WIP: VM-based test harness (vm_builder)

zplizzi:
## Status: WIP — VM build does not finish

Boots Ubuntu 24.04 in QEMU, seeds it with cloud-init, then runs `ansible/setup.yml` against it to produce a reusable qcow2. Goal is a deterministic local target for integration tests so they don't need a real server.

## What works

- QEMU boot (arm64 on Apple Silicon, x86_64 on Linux), cloud-init SSH key injection.
- `ansible/setup.yml` runs through ~33 of 34 tasks: user setup, apt, podman, dev env, code sync, pixi install.
- Build artifact cached at `tests/.vm-cache/built.qcow2`, keyed on a SHA over every git-tracked file under `ansible/`.

## What still needs done

See line comments below.

1. **Pixi platforms — primary blocker.** `tools.pixi.workspace.platforms = ["linux-64"]`, but on Apple Silicon vm_builder builds an arm64 VM, so `pixi install` fails with `unsupported-platform: linux-aarch64`. Options: force x86_64 TCG VM (slow, matches prod), or add `linux-aarch64` to platforms (faster, diverges from prod arch). Comment in `pyproject.toml` notes coredns "doesn't work on arm mac" — that's about osx-arm64; linux-aarch64 conda builds do exist.
2. **rsync prerequisite.** Apple's `openrsync` 2.6.9 is incompatible with Linux rsync 3.x. Build needs `brew install rsync` on the dev machine. Should be added to a doctor check or README.
3. **Early-exit detection.** `_wait_ssh` blindly waits up to 600s; if QEMU exits at startup, the script hangs the full timeout instead of failing immediately. Should poll `qemu.poll()` in the loop.
4. **No actual tests yet.** `boot_vm_image()` context manager is defined but no test file uses it. Once the build works, wire up a smoke test.
5. **`ansible/secrets/certbot_private_key.json` is required for prod deploys** (sync_code.yml ACME copy). I added a tag so the VM build can skip it, but contributors running `setup.yml` directly without the file will still fail. Consider documenting or detecting absence.

## Repro

```sh
brew install qemu rsync ansible    # rsync 3.x needed
uv run python tests/vm_builder.py
```
Currently fails at `TASK [Install pixi dependencies (caddy, coredns)]` on Apple Silicon.

### DIFF
--- a/ansible/tasks/sync_code.yml
+++ b/ansible/tasks/sync_code.yml
@@ -19,3 +19,5 @@
     owner: host
     group: host
     mode: "0600"
+  tags:
+    - acme_key
--- a/tests/vm_builder.py
+++ b/tests/vm_builder.py
@@ -1,0 +1,384 @@
+"""Build a local test VM image.
+
+Boots Ubuntu 24.04 in QEMU, runs ansible/setup.yml against it, and saves the
+resulting qcow2. Skips rebuild if no git-tracked file under ansible/ has
+changed since the last successful build.
+
+mac: requires `brew install qemu`. apple silicon also uses the edk2 firmware
+that ships with brew qemu. linux: requires qemu, kvm, genisoimage.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import hashlib
+import os
+import platform
+import shutil
+import socket
+import subprocess
+import sys
+import time
+from dataclasses import dataclass
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+ANSIBLE_DIR = REPO_ROOT / "ansible"
+CACHE_DIR = Path(__file__).resolve().parent / ".vm-cache"
+
+UBUNTU_BASE_URL = "https://cloud-images.ubuntu.com/noble/current"
+UBUNTU_IMG = {
+    "amd64": "noble-server-cloudimg-amd64.img",
+    "arm64": "noble-server-cloudimg-arm64.img",
+}
+
+DISK_SIZE = "20G"
+RAM_MB = 4096
+CPUS = 4
+SSH_PORT = 2222
+SSH_TIMEOUT_S = 600
+
+
+def _arch() -> str:
+    m = platform.machine().lower()
+    if m in ("x86_64", "amd64"):
+        return "amd64"
+    if m in ("arm64", "aarch64"):
+        return "arm64"
+    raise RuntimeError(f"unsupported arch: {m}")
+
+
+def _require(cmd: str) -> str:
+    p = shutil.which(cmd)
+    if not p:
+        raise RuntimeError(f"required command not found: {cmd!r}. on mac: brew install qemu ansible")
+    return p
+
+
+def _ansible_hash() -> str:
+    """SHA-256 over the content of every git-tracked file under ansible/."""
+    out = subprocess.check_output(["git", "-C", str(REPO_ROOT), "ls-files", "ansible"], text=True)
+    h = hashlib.sha256()
+    for rel in sorted(out.splitlines()):
+        p = REPO_ROOT / rel
+        if not p.is_file():
+            continue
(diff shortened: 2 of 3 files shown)


tests/vm_builder.py
```diff
@@ -0,0 +1,390 @@
...
+SSH_TIMEOUT_S = 600
+
+
+def _arch() -> str:
+    m = platform.machine().lower()
+    if m in ("x86_64", "amd64"):
+        return "amd64"
+    if m in ("arm64", "aarch64"):
+        return "arm64"
+    raise RuntimeError(f"unsupported arch: {m}")
```
zplizzi, line 48: **TODO (primary blocker):** on Apple Silicon this returns `arm64`, so we boot an arm64 VM and `pixi install` fails with `unsupported-platform: linux-aarch64` (pixi workspace is locked to `linux-64`).

Two fixes:
1. Force x86_64 here on darwin (TCG emulation, slow but matches prod).
2. Add `linux-aarch64` to `tool.pixi.workspace.platforms` in pyproject.toml. Conda has caddy/coredns aarch64 builds; the existing arm64 caveat in pyproject.toml is about *osx*-arm64, not linux-aarch64.

tests/vm_builder.py
```diff
@@ -0,0 +1,390 @@
...
+            _ssh_cmd(key) + ["true"],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
+            text=True,
+        )
+        if r.returncode == 0:
+            return
+        last_err = r.stderr.strip()
+        time.sleep(3)
+    raise TimeoutError(f"VM not SSH-reachable after {timeout}s; last: {last_err}")
```
zplizzi, line 251: **TODO:** if QEMU exits at startup (e.g., port-forward setup race, missing firmware), this still waits the full 600s before failing. Should poll `qemu.poll()` each iteration and fail fast with the serial.log tail.

tests/vm_builder.py
```diff
@@ -0,0 +1,390 @@
+"""Build a local test VM image.
+
+Boots Ubuntu 24.04 in QEMU, runs ansible/setup.yml against it, and saves the
+resulting qcow2. Skips rebuild if no git-tracked file under ansible/ has
+changed since the last successful build.
+
+mac: requires `brew install qemu`. apple silicon also uses the edk2 firmware
+that ships with brew qemu. linux: requires qemu, kvm, genisoimage.
```
zplizzi, line 8: **TODO:** also requires `brew install rsync` (3.x). Apple's `/usr/bin/rsync` is openrsync 2.6.9 and ansible's synchronize module fails against Linux rsync 3.x with `mkpath: Operation not supported`. Either document this here or add to a doctor check.

ansible/tasks/sync_code.yml
```diff
@@ -19,3 +19,5 @@
     owner: host
     group: host
     mode: "0600"
+  tags:
+    - acme_key
```
zplizzi, line 23: **TODO (followup):** running `ansible/setup.yml` on a fresh dev machine without `secrets/certbot_private_key.json` will hard-fail here. Either auto-skip when the file is missing (with a warning) or document the prereq.

tests/vm_builder.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+@dataclass
+class RunningVM:
+    ssh_port: int
+    ssh_key: Path
+    overlay: Path
+    _qemu: subprocess.Popen
+
+    def ssh(self, *args: str, check: bool = True) -> subprocess.CompletedProcess:
```
zplizzi, line 343: **TODO:** `boot_vm_image()` is defined but not yet used by any test. Once the build works end-to-end, wire up a smoke test (e.g., assert openhost.service is active in the ephemeral overlay).

PR #50, opened 2026-04-29
Zack/new test layout

zplizzi:
moves tests into package namespaces, like `compute_space.tests`. this is considered somewhat "bad practice" because tests get shipped w the code (unless explicitly excluded), and you can do sloppy stuff like importing test code from main code. but the alternatives for multi-package repos are quite bad and very confusing; this layout at least makes sense and works without jankiness. i prefer simple+understandable over magic, and i don't really see why this layout is so bad.

also moves to a "src" layout where code is at `name/src/name`. this is standard best practice nowadays, for reasons.

previously i was having a lot of issues with test imports colliding and such; the old setup was non-tenable. it was just a question of what the new setup would be.


zplizzi:
also it turned out the container tests weren't actually adding the containers flag, so wasn't running the container tests

andrewlaack (approved):
👍

PR #53, opened 2026-04-30
Add app_archive storage tier with optional JuiceFS-on-S3 backing

andrewlaack:
Adds a third per-app storage tier alongside app_data and app_temp_data: app_archive, opt-in via [data] app_archive=true in the manifest, bind-mounted into containers at /data/app_archive/<app>/. The host-side backing is operator-selected: defaults to a local-disk subdirectory under persistent_data_dir (so existing zones get the tier with no operator action), and can be repointed at a JuiceFS-on-S3 mount via an ansible role gated behind juicefs_enabled.

Apps see the same in-container path either way; only the host-side storage changes.

Phase 1+2 (the in-process plumbing): manifest field, validator, provision_data + run_container + deprovision_data signature changes, rename_app extension. Phase 3 (ansible): JuiceFS install/format/mount role with systemd ordering, daily metadata-dump timer for restic-based backup, operator docs.

Verified end-to-end on a real EC2 VM that an opt-in app sees /data/app_archive/<name>/ as expected and that backwards-compat is intact for apps without the opt-in. The JuiceFS-on-S3 path was NOT end-to-end tested against a live bucket — the role and its tests cover the static config rendering, sentinel-guarded format idempotency, systemd unit ordering (Requires=, not just After=), and credential handling (env file, never argv), but a fresh JuiceFS-backed VM with a real S3 round-trip still wants exercising before any production rollout. Happy to drive that test if you want; it needs a small vm-manager extension to plumb the juicefs_* ansible vars through /api/create.

Manifest validator rejects app_archive=true without app_data (or sqlite, or access_all_data), since the archive tier may be on a network FS and SQLite/advisory-lock-using apps need somewhere safe for working state.

344 unit tests pass, including new coverage for: provision/deprovision happy + missing-archive-dir + access_all_data variants, ansible template rendering (config.toml, openhost.service, juicefs-mount, juicefs-meta-dump), rename_app three-tier rename + rollback + rollback-itself-fails. Six rounds of vet across codex and claude harnesses, all findings addressed.

Companion fix already merged: openhost-vm-manager#16 threads the branch parameter through to ansible as openhost_commit. Without that, /api/create branch=X silently fell back to main on the deployed VM.

### DIFF
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -47,4 +47,9 @@
   tasks:
     - import_tasks: tasks/sync_code.yml
     - import_tasks: tasks/pixi.yml
+    # JuiceFS must be installed + mounted BEFORE the openhost service
+    # is enabled, since openhost.service has Requires=juicefs-mount.service
+    # when juicefs_enabled is true.
+    - import_tasks: tasks/juicefs.yml
+      when: juicefs_enabled | default(false) | bool
     - import_tasks: tasks/setup_openhost_service.yml
--- a/ansible/tasks/juicefs.yml
+++ b/ansible/tasks/juicefs.yml
@@ -106,6 +106,18 @@
     group: host
     mode: "0750"
 
+# /var/log is root-owned, so the host user can't write juicefs's
+# mount log there directly.  Pre-create a dedicated subdir owned
+# by host so the systemd unit can write without sudo.
+- name: Create juicefs log dir
+  become: yes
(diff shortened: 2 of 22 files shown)


andrewlaack:
Wrong shape — archive backend should be configurable from the dashboard, not at provisioning time. Branch kept for reference; opening a fresh PR with the dashboard-driven design.

PR #58, opened 2026-05-01
Archive data tier (S3 via JuiceFS), disabled by default

andrewlaack:
Adds an archive data tier apps opt into via [data] app_archive = true, backed by S3 via JuiceFS. Operator configures S3 once via the System tab; fresh zones default to disabled. Once configured, the choice is permanent for that zone.

One new migration (v5), one new core module (core/archive_backend.py), system-tab UX for one-shot configure, JuiceFS layout under <openhost_data>/juicefs/state|runtime/. Per-zone S3 isolation via an operator-supplied s3_prefix that maps to the JuiceFS volume name.

Apps with app_archive=true require backend='s3' to install/reload/rename. access_all_data semantics relaxed: gets the archive when it exists, skipped when it doesn't.

394 unit tests pass; e2e ec2/gcp, test-containers, pre-commit all green. Live-verified on andrew-3 against s3://imbue-openhost/andrew-3/.

Supersedes PR 54.


andrewlaack:
There are three migrations because I added them incrementally as I worked, but they could reasonably be compressed into a singular migration; not sure if it matters.

compute_space/src/compute_space/core/archive_backend.py
```diff
@@ -0,0 +1,1032 @@
...
+
+def _juicefs_binary(config: Config) -> str:
+    return os.path.join(_juicefs_install_dir(config), f"juicefs-{JUICEFS_VERSION}")
+
+
+def _juicefs_meta_db(config: Config) -> str:
+    return os.path.join(_juicefs_state_dir(config), "meta.db")
+
+
+_LEGACY_META_DB_BASENAME = "juicefs-meta.db"
```
zplizzi: let's get rid of this and _migrate_legacy_layout if you're the only person that needed migrated?

compute_space/src/compute_space/core/archive_backend.py
```diff
@@ -0,0 +1,1032 @@
...
+import attr
+import boto3
+import botocore.exceptions  # noqa: F401  -- imported for ``except`` matching downstream
+
+from compute_space.config import Config
+from compute_space.core.logging import logger
+
+# Pin a specific JuiceFS release; sha256 is verified before extract so a
+# compromised release page can't swap the tarball.
+JUICEFS_VERSION = "1.3.1"
```
zplizzi, line 29: interestingly nix has a package for juicefs, conda doesn't. +1 for nix. https://mynixos.com/nixpkgs/package/juicefs

compute_space/src/compute_space/core/archive_backend.py
```diff
@@ -0,0 +1,1032 @@
...
+
+def juicefs_state_dir(config: Config) -> str:
+    return _juicefs_state_dir(config)
+
+
+def is_juicefs_installed(config: Config) -> bool:
+    return os.path.isfile(_juicefs_binary(config)) and os.access(_juicefs_binary(config), os.X_OK)
+
+
+def install_juicefs(config: Config) -> None:
```
zplizzi, line 83: (goes without saying that i don't like this, but i don't see a better option with how we have things setup now)

compute_space/src/compute_space/core/archive_backend.py
```diff
@@ -0,0 +1,1032 @@
...
+    db.commit()
+
+
+# Anchored on TOML key=value shape so substring matching can't false-match
+# ``app_archive = false`` alongside an unrelated ``= true``.
+_MANIFEST_REQUIRES_ARCHIVE_RE = re.compile(r"(?m)^\s*app_archive\s*=\s*[Tt][Rr][Uu][Ee]\b")
+_MANIFEST_USES_ARCHIVE_RE = re.compile(r"(?m)^\s*(?:app_archive|access_all_data)\s*=\s*[Tt][Rr][Uu][Ee]\b")
+
+
+def manifest_requires_archive(manifest_raw: str) -> bool:
```
zplizzi, line 393: this is for openhost.toml? why can't we just use the Config object? i'd rather not be parsing it in multiple places.

compute_space/src/compute_space/core/manifest.py
```diff
@@ -296,6 +297,15 @@ def parse_manifest_from_string(raw_text: str) -> AppManifest:
             app_name,
         )
 
+    has_local_tier = data_section.get("app_data") or data_section.get("sqlite") or data_section.get("access_all_data")
```
zplizzi: idk if we need this check? feels paternalistic.

compute_space/src/compute_space/core/startup.py
```diff
@@ -123,6 +124,13 @@ def init_app(app: Quart) -> None:
     """Initialize DB and app state. Call after data directories are ready."""
     config = app.openhost_config  # type: ignore[attr-defined]
     init_db(app)
+    db = sqlite3.connect(config.db_path)
+    try:
+        new_config = archive_backend.attach_on_startup(config, db)
+    finally:
+        db.close()
+    app.openhost_config = new_config  # type: ignore[attr-defined]
```
zplizzi: can you arrange this to be like `config = attr.evolve(config, arg_to_update=new_value)`, in this file? the current structure makes it hard to see what's getting mutated in config without going and reading very unrelated code.

compute_space/src/compute_space/db/versioned/migrations/v0005_archive_backend.py
```diff
@@ -0,0 +1,17 @@
+"""v5: add the ``archive_backend`` table.
```
zplizzi: yeah let's merge these into 1 if possible

compute_space/src/compute_space/web/routes/api/apps.py
```diff
@@ -161,6 +167,32 @@ async def api_add_app() -> ResponseReturnValue:
...
+    if manifest.app_archive:
+        backend_state = archive_backend.read_state(db)
+        if backend_state.backend == "disabled":
+            shutil.rmtree(clone_dir, ignore_errors=True)
+            return jsonify(
+                {
+                    "error": "This app uses the app_archive data tier, but "
+                    "the archive backend has not been configured on "
+                    "this zone.  Visit the System page and pick an "
+                    "archive backend (local disk or S3) before "
```
zplizzi: why not default to local disk, so we never hit this path?

compute_space/src/compute_space/web/routes/api/apps.py
```diff
@@ -280,6 +312,17 @@ async def reload_app(app_name: str) -> ResponseReturnValue:
     if _is_removing(app_row):
         return jsonify({"error": "App is being removed"}), 409
 
+    if not archive_backend.is_archive_dir_healthy(config, db):
+        if _manifest_requires_archive(app_row["manifest_raw"] or ""):
```
zplizzi: can we parse the manifest properly vs using a regex here? there should be only one parsing function.

compute_space/src/compute_space/core/archive_backend.py
```diff
@@ -0,0 +1,1032 @@
...
+    try:
+        client = _s3_client(s3_region, s3_endpoint, s3_access_key_id, s3_secret_access_key)
+        client.head_bucket(Bucket=s3_bucket)
+    except Exception as exc:
+        return f"S3 reachability test failed: {exc}"
+    return None
+
+
+@attr.s(auto_attribs=True, frozen=True)
+class AppHook:
```
zplizzi: hmm is there no better way to do this? i'd rather do some monkeypatching/magic in the tests than pollute the main code with testing utilities.

compute_space/src/compute_space/web/routes/api/archive_backend.py
```diff
@@ -0,0 +1,263 @@
...
+            "no leading/trailing dash) — it doubles as the JuiceFS volume name "
+            "and so has to satisfy JuiceFS's name regex.  For multi-zone "
+            "isolation under a shared bucket, give each zone a unique "
+            "single-segment name (e.g. ``andrew-3`` for zone andrew-3, "
+            f"``andrew-1`` for zone andrew-1).  Got: {cleaned!r}"
+        )
+    return cleaned
+
+
+def _build_hook(app: Any) -> AppHook:  # ``Any`` to avoid import-cycle on the Quart app type
```
zplizzi: same comment as before

compute_space/src/compute_space/core/startup.py
zplizzi: actually, i don't think this should live in the config at all. config should be only things that are settable by the user, not things set at runtime imo. i think we can just have a `archive_backend.archive_folder_path()`  function that returns the appropriate path? it can reference a global or just check if juicefs is mounted or whatever.

zplizzi (approved):
lgtm overall, will dm about some other thoughts

PR #60, opened 2026-05-05
Owner username: optional setup field + OPENHOST_OWNER_USERNAME env var

andrewlaack:
Per Zack's suggestion: rather than build a multi-account abstraction, plumb the owner's chosen display name through to apps via env var so SSO consumers (PeerTube, Mastodon, etc.) can seed account names instead of inventing defaults the user can't rename.

Three user-facing surfaces, all driven from the existing owner.username column (no schema migration):

- /setup form: optional Username field above password. Blank falls back to the legacy 'owner' default; validation accepts [A-Za-z0-9][A-Za-z0-9._-]{0,49}.
- /settings page: editor for the same value, matching the Git Remote pattern (text input + Save). Sign-in stays password-only.
- Per-app containers: provision_data stamps OPENHOST_OWNER_USERNAME on every app whose owner has a name configured; omitted entirely on pre-setup zones so apps can distinguish 'no name' from a default.

Three derived fixes that fall out of making 'owner' configurable:

- Setup + login JWTs sign with the persisted owner.username instead of a hardcoded 'owner' literal, so the proxy's X-OpenHost-Is-Owner check still works.
- Token-refresh middleware reads the current owner.username on refresh instead of recycling the stale claim, so a mid-session username change propagates.
- _validate_api_token + middleware refresh both refuse to mint claims when no owner row exists, so leftover cookies from a wiped instance can't bootstrap into a freshly-claimed zone.

47 new tests in test_owner_username.py covering validator boundary, read/update helpers, env-var stamping, GET/POST settings routes, /setup persistence + JWT claim, /login JWT claim, middleware refresh (change-pickup + orphan-cookie rejection), _validate_api_token (populated + missing owner), and proxy._identity_headers. Documented in creating_an_app.md and manifest_spec.md.

Draft. Not deployed. Single-owner; explicitly does not introduce a multi-account abstraction.


compute_space/src/compute_space/core/auth/auth.py
```diff
@@ -1,9 +1,105 @@
 import hashlib
+import re
+import sqlite3
 from datetime import UTC
 from datetime import datetime
 
 from compute_space.db import get_db
 
+# starting with an alphanumeric.  Length cap matches Mastodon's 30-
```
zplizzi: nit: reflow these comments to 119 chars. i think i have a note in claude.md or style_guide.md about this, but maybe your setup doesn't read those.

compute_space/src/compute_space/core/auth/auth.py
```diff
@@ -1,9 +1,105 @@
...
+# char column with headroom; leading-alphanumeric avoids the
+# common SSO pitfall of leading-dot identifiers.
+#
+# Note we do NOT enforce the strictest downstream constraint
+# (PeerTube's ``[a-z0-9._]{1,50}`` — lowercase only, no hyphen).
+# Forcing lowercase or dropping ``-`` would prevent legitimate
+# identifiers like ``Andrew`` or ``zack-2``; instead, downstream
```
zplizzi: i would be inclined to only allow lowercase - we may want to share this with subdomain name for hosted spaces (zack.selfhost.imbue.com or whatever), and idk why we really need to support caps. it's also always easier to loosen such rules in the future than to tighten.

zplizzi: 50 chars also feels like a lot, idk

compute_space/src/compute_space/core/auth/auth.py
```diff
@@ -1,9 +1,105 @@
...
+    if not _OWNER_USERNAME_RE.match(value):
+        return "Username must start with a letter or digit and contain only letters, digits, ``.``, ``_`` or ``-``."
+    return None
+
+
+def read_owner_username(db: sqlite3.Connection) -> str | None:
+    """Return the configured owner username, or ``None`` if no owner
+    row exists yet (pre-setup) or the column is empty.
+
+    Used by the per-app provisioning path to stamp
```
zplizzi: nit: these comments are kinda unnecessarily verbose

compute_space/src/compute_space/core/data.py
```diff
@@ -81,13 +81,22 @@ def provision_data(
...
+    owner_username: str | None = None,
 ) -> dict[str, str]:
     """Create data directories for an app based on manifest permissions.
     Returns a dict of environment variable name -> value.
 
     Apps only get filesystem access to directories they explicitly request
     via app_data, app_temp_data, and app_archive flags in [data].  SQLite
     entries implicitly enable app_data.
+
+    ``owner_username``, when set, is forwarded as
```
zplizzi: i think we should just always set this? it's fine if it's `owner` for existing spaces.

compute_space/src/compute_space/web/routes/pages/login.py
```diff
@@ -54,7 +54,10 @@ async def login() -> ResponseReturnValue:
     if not bcrypt.checkpw(password.encode(), owner["password_hash"].encode()):
         return await render_template("login.html", error="Invalid password")
 
-    access_token = create_access_token("owner")
+    # Invariant: the JWT's ``sub``/``username`` claim matches the
```
zplizzi: again comments don't seem like they add much

compute_space/src/compute_space/web/routes/pages/setup.py
```diff
@@ -85,20 +87,52 @@ async def setup() -> ResponseReturnValue:
 
     password = form.get("password", "")
     confirm = form.get("confirm_password", "")
+    # Username is optional; blank submissions fall back to
+    # ``DEFAULT_OWNER_USERNAME`` so a hurried operator can complete
+    # setup without thinking about names, and zones with no
+    # configured username still produce a working SSO claim.  The
+    # operator can change it later from the settings page.
+    username_raw = form.get("username", "").strip()
```
zplizzi: nit: save a couple lines by putting DEFAULT_OWNER_USERNAME here

compute_space/src/compute_space/web/routes/proxy.py
```diff
@@ -22,7 +24,26 @@
...
 
 def _identity_headers(claims: dict[str, str] | None) -> dict[str, str]:
-    if claims and claims.get("sub") == "owner":
+    """Build the per-request identity headers forwarded to apps.
+
+    Emits exactly ``X-OpenHost-Is-Owner: true`` when the bearer's
+    ``sub`` claim matches the persisted ``owner.username``; omits
+    the header otherwise.
+
+    Every successfully-validated JWT in this codebase today belongs
```
zplizzi: drop comments about migration from code

docs/src/manifest_spec.md
```diff
@@ -87,6 +87,7 @@ The host provisions requested data services and injects connection info as envir
 - `OPENHOST_APP_ARCHIVE_DIR` — `/data/app_archive/{app_name}` (only if app_archive access granted)
 - `OPENHOST_AUTH_PUBLIC_KEY` — PEM-encoded JWT public key for token verification (only if signing keys are available)
 - `OPENHOST_ROUTER_URL` — URL of the router's HTTP server, reachable from inside the container.
+- `OPENHOST_OWNER_USERNAME` — the compute space owner's chosen display name; use to seed SSO account names. Omitted when the owner has not configured one.
```
zplizzi: again, make this always set

docs/src/creating_an_app.md
```diff
@@ -113,6 +113,7 @@ The router injects these environment variables into your app:
 | `OPENHOST_APP_DATA_DIR` | `/data/app_data/my-app` | Path to the app's persistent data directory. Set when `app_data`, `sqlite`, or `access_all_data` is requested   |
 | `OPENHOST_APP_TEMP_DIR` | `/data/app_temp_data/my-app` | Path to the app's temporary data directory. Set when `app_temp_data` or `access_all_data` is requested          |
 | `OPENHOST_SQLITE_<NAME>` | `/data/app_data/my-app/sqlite/main.db` (for `sqlite = ["main"]`) | Path to a provisioned SQLite database file. Set once per entry in `sqlite`                                      |
+| `OPENHOST_OWNER_USERNAME` | `alice` | The compute space owner's chosen display name. Use this to seed account names for SSO integrations (PeerTube, Mastodon, etc.) so the owner doesn't get a generated default they can't change. Omitted entirely when the owner has no username configured (pre-setup zones); apps should treat the absent variable as "no name available" rather than substituting a literal default. |
```
zplizzi: again, make this always set

zplizzi: we should never be starting an app if the zone is pre-setup?

zplizzi (approved):
various nits but overall lgtm

PR #73, opened 2026-05-11
remove path-based app routing (zone.com/app/route), cleanup the app proxy routing a bit

zplizzi:


compute_space/src/compute_space/web/routes/identity.py
```diff
@@ -0,0 +1,105 @@
+import base64
```
zplizzi, line 1: this is all unchanged, just refactored out of the proxy file

compute_space/src/compute_space/core/apps.py
```diff
@@ -815,3 +817,40 @@ def remove_app_background(app_name: str, keep_data: bool, config: Config) -> Non
             logger.exception("Could not record removal failure for %s", app_name)
     finally:
         db.close()
+
+
+def parse_app_from_host(host: str) -> str | None:
```
zplizzi, line 822: these 3 funcs are unchanged, just moved

PR #81, opened 2026-05-13
Add single-command server provisioning script

andrewlaack:
Adds scripts/provision.sh for bootstrapping a fresh Ubuntu 24.04 server into a running OpenHost instance with one command:

  curl -fsSL .../provision.sh | bash -s -- --domain myhost.example.com

The script creates the host user, installs ansible+git, clones the repo, and runs a new ansible/local_setup.yml playbook that reuses all existing task files (apt_packages, containers, dev_machine, pixi, setup_openhost_service) via local connection mode. ACME account key is auto-generated if missing.

No code duplication -- local_setup.yml imports the same tasks as setup.yml, so changes to ansible tasks apply to both remote and local provisioning automatically.

Needs testing on a fresh VM (couldn't SSH into vm-manager instances from agent-host).

### DIFF
--- a/ansible/local_setup.yml
+++ b/ansible/local_setup.yml
@@ -6,6 +6,13 @@
 #
 # Usage (normally called by provision.sh, not directly):
 #   ansible-playbook ansible/local_setup.yml -e domain=<domain> -e public_ip=<ip> --connection=local -i localhost,
+#
+# Privilege model: provision.sh runs as root. In setup.yml, the "Setup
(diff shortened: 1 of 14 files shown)


andrewlaack:
on a fully provisioned vps running the correct version of ubuntu, with pre-configured dns routing, there should be a singular script that can run on the vps to fully provision it, instead of requiring an external host to run an ansible-playbook.

andrewlaack:
Superseded by #85, #86, and #87 which split this into three focused PRs.

PR #88, opened 2026-05-13
security(proxy): strip zone_auth cookie and inbound X-OpenHost-* headers

kilo-gi:
The proxy was forwarding the owner's zone_auth JWT cookie to every app backend. Apps run on subdomains of zone_domain and the cookie is set with Domain=zone_domain, so any app the owner visits could read the JWT from its incoming request and replay it against compute_space admin APIs or any other app — a privilege-escalation vector for a platform that hosts untrusted apps.

Separately, X-OpenHost-* headers were not stripped from inbound requests. Clients (including public-path callers) could spoof X-OpenHost-Is-Owner: true directly to apps. Now the router is the sole authority for these headers.

### DIFF
--- a/compute_space/src/compute_space/tests/test_integration.py
+++ b/compute_space/src/compute_space/tests/test_integration.py
@@ -1134,6 +1134,30 @@ class TestContainerE2E:
         assert headers.get("X-Forwarded-Proto") != "evil"
         assert headers.get("X-Forwarded-Host") != "evil.example.com"
 
+    def test_proxy_strips_zone_auth_cookies(self, admin_session, config):
+        """The owner's zone_auth / zone_refresh cookies must not reach the backend app."""
+        r = admin_session.get(f"{_app_url(config, 'test-app')}/echo-headers")
+        assert r.status_code == 200
+        cookie_header = r.json()["headers"].get("Cookie", "")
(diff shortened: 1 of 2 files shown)


zplizzi (changes requested):
good catches. can you factor some of this into a common helper so we're duplicating less code between the websocket and regular proxy, if that's doable?

kilo-gi:
Followup on the duplication feedback: factored the shared filtering into a single `_sanitize_forwarded_headers` helper used by both `proxy_request` and `ws_proxy`. The HTTP and WS paths still each spell out their own protocol-level hop-by-hop excludes (they differ — WS drops `upgrade`/`sec-websocket-*`, HTTP drops `accept-encoding`), but the security policy (strip zone auth cookies, drop inbound X-OpenHost-*) now lives in one named place.

One side-effect to flag: `proxy_request` previously stripped `Cookie` from the forwarded headers and passed the remaining cookies via httpx's `cookies=` kwarg. To unify with the WS path (which has no equivalent), it now filters the raw `Cookie` header and forwards it as-is. Net effect for the backend is the same set of cookies arriving, but their original encoding is preserved instead of being re-encoded by httpx — arguably more correct for a transparent proxy.

zplizzi (approved):

PR #90, opened 2026-05-14
Zack/remove quart from core

zplizzi:
and a bit of other minor restructuring. putting all routes actually in the routes/ dir. should be no functional changes.


zplizzi:
cc @kilo-gi a bit more restructuring of the auth code

PR #91, opened 2026-05-14
Add deployment reference skill

andrewlaack:
Adds a purely informational skill document at skills/deployment/SKILL.md that describes the OpenHost platform and how apps are deployed to it.

Covers: platform architecture, network stack, app manifest (openhost.toml) schema, environment variables, data storage tiers, authentication/routing model, app lifecycle, CLI commands, HTTP API endpoints, rootless podman constraints, infrastructure deployment, and example manifests.

The skill is informational only -- it describes how things work, it does not tell agents what to do.

### DIFF
--- a/skills/deployment/SKILL.md
+++ b/skills/deployment/SKILL.md
@@ -1,0 +1,427 @@
+---
+name: openhost-deployment
+description: Reference information about the OpenHost platform, its architecture, and how apps are deployed and managed. This is a purely informational skill -- it describes how things work, not what to do.
+---
+
+# OpenHost Deployment Reference
+
+This document describes the OpenHost platform, its architecture, and how
+applications are deployed and managed on it. It is intended as a reference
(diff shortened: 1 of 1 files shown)


skills/deployment/SKILL.md
```diff
@@ -0,0 +1,134 @@
...
+  -i <IP>, \
+  -e domain=<domain> \
+  --private-key=~/.ssh/YOUR_KEY
+```
+
+The trailing comma after the IP is required (tells ansible it's a host list, not a file).
+
+To deploy a specific commit: `-e openhost_commit=$(git rev-parse HEAD)`. The commit must be pushed.
+
+## Option 3: vm-manager
```
andrewlaack: not sure if we want this part; this is exceptionally useful for internal stuff bc this is what we should be using, but sort of complicates distribution to end users.

PR #97, opened 2026-05-15
Add change password to settings

andrewlaack:
Add a change password form to the settings page with current password verification, confirmation matching, and minimum length validation.


compute_space/src/compute_space/web/routes/api/settings.py
```diff
@@ -211,6 +212,48 @@ async def restart_compute_space() -> None:
...
+    if new_pw != confirm:
+        raise HTTPException(detail="Passwords do not match", status_code=400)
+    if len(new_pw) < 8:
+        raise HTTPException(detail="Password must be at least 8 characters", status_code=400)
+
+    row = db.execute("SELECT user_id, password_hash FROM users LIMIT 1").fetchone()
+    if not row:
+        raise HTTPException(detail="No owner found", status_code=404)
+
+    if not bcrypt.checkpw(current.encode(), row["password_hash"].encode()):
```
zplizzi, line 244: i'd rather use `validate_password` from core - there should only be one place where this key auth logic happens.

compute_space/src/compute_space/web/routes/api/settings.py
```diff
@@ -211,6 +212,48 @@ async def restart_compute_space() -> None:
...
+        raise HTTPException(detail="Current password is incorrect", status_code=403)
+
+    new_hash = bcrypt.hashpw(new_pw.encode(), bcrypt.gensalt()).decode()
```
zplizzi, line 247: nit: move the auth logic around changing the pw into core/auth.py - i prefer to have all the auth logic (and logic in general) centralized so the routes are just dumb-ish interfaces. keeps it easier to review in the future.

PR #109, opened 2026-05-20
Migrate api/system to Litestar + add fallback wrapper

zplizzi:
## Summary

Stacked on #108. Third in the series. Moves the 12 ``/api/...`` system routes (tokens CRUD, compute_space_logs, health, security audit, listening ports, storage status / guard toggle, SSH status / toggle, drop docker cache, restart router) to native Litestar handlers with attrs-typed request and response bodies. POST endpoints take typed JSON bodies (no more ``form.get(key)`` lookups); dashboard JS for ``api_tokens_create`` and ``toggle_storage_guard`` switches from ``FormData`` to ``JSON.stringify`` accordingly.

## Notable: outer fallback wrapper replaces the @asgi mount

Litestar 2.x routes path-parameterised handlers (``/api/tokens/{token_id:int}``, ``/app_detail/{app_id}``, ``/api/services/v2/call/...``) at lower precedence than a mount at ``/``, so ``@asgi(path="/", is_mount=True)`` was silently shadowing every Litestar route that had a path param — including PR #107's ``/app_detail/{app_id}`` and this PR's ``DELETE /api/tokens/{id}``. Specific routes never ran; every such request 404'd out of the Quart fallback instead.

Replaced the mount with a middleware-style wrapper that invokes Litestar first and only delegates to Quart on a routing 404. Specific routes win, the Quart sub-app still catches everything else, and the request body is left untouched on the 404 path so Quart can re-read it.

## Other fixups in this PR

- pre-existing typo fixed: ``init_db(_FakeApp(cfg.db_path))`` → ``init_db(cfg.db_path)`` in test_remove_app_route.py (an undefined ``_FakeApp`` snuck in via #103).
- integration test ``test_api_token_create_and_use`` now asserts 204 on DELETE (Litestar's default for ``@delete``) and sends ``json=`` instead of ``data=`` for the create-token POST.
- ``test_build_cache.py`` rewritten to drive the route via Litestar TestClient.

## Test plan

- [x] `pixi run -e dev pytest compute_space/src/compute_space/tests/` — 501 passing, 32 skipped, 0 failures
- [x] `pixi run -e dev mypy compute_space/src/compute_space/` — clean
- [x] `pixi run -e dev ruff check compute_space/src/compute_space/` — clean
- [ ] Manual: tokens UI on the dashboard (create + delete), storage guard pause/resume on /system, SSH toggle, drop docker cache

### DIFF
--- a/compute_space/src/compute_space/web/routes/api/system.py
+++ b/compute_space/src/compute_space/web/routes/api/system.py
@@ -49,8 +49,8 @@ class CreateTokenRequest:
     sentinel ``"never"`` (= no expiry) and a numeric value share one field
     without forcing the JS client to send different keys."""
 
-    name: str = ""
-    expiry_hours: str = ""
+    name: str
+    expiry_hours: str | None = None
 
 
 @attr.s(auto_attribs=True, frozen=True)
@@ -135,9 +135,9 @@ async def api_tokens_create(
     data: CreateTokenRequest, db: sqlite3.Connection
 ) -> Response[CreatedToken] | Response[ErrorResponse]:
     name = data.name.strip() or "Untitled"
-    expiry_hours_raw = data.expiry_hours.strip()
+    expiry_hours_raw = data.expiry_hours.strip() if data.expiry_hours else ""
     expires_at: datetime | None
-    if expiry_hours_raw == "" or expiry_hours_raw.lower() == "never":
+    if not expiry_hours_raw or expiry_hours_raw.lower() == "never":
         expires_at = None
     else:
         try:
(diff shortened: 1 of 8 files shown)


compute_space/src/compute_space/web/routes/api/system.py
```diff
@@ -21,210 +26,243 @@
...
+    expired: bool
+
+
+@attr.s(auto_attribs=True, frozen=True)
+class CreateTokenRequest:
+    """Body for ``POST /api/tokens``.  ``expiry_hours`` is a string so the
+    sentinel ``"never"`` (= no expiry) and a numeric value share one field
+    without forcing the JS client to send different keys."""
+
+    name: str = ""
```
zplizzi: make required

compute_space/src/compute_space/web/routes/api/system.py
```diff
@@ -21,210 +26,243 @@
...
+    expiry_hours: str = ""
```
zplizzi: make default None.

PR #124, opened 2026-05-21
Bind router to loopback by default to prevent cleartext credential exposure

zplizzi:
## Summary

- **Security fix**: Change `DefaultConfig.host` from `"0.0.0.0"` to `"127.0.0.1"` in `compute_space/src/compute_space/config.py`
- The router (Hypercorn) was listening on all interfaces by default, making port 8080 directly reachable from the internet over plaintext HTTP -- bypassing Caddy's TLS termination
- This exposed login passwords (served and submitted over cleartext) and Bearer API tokens to network eavesdroppers
- Caddy connects to the router via `localhost:8080`, so loopback binding is sufficient
- Deployments that explicitly set `host` in their config (Ansible template, `config_gen.py`) are unaffected

## Test plan

- [x] All 499 compute_space unit tests pass (32 skipped for container/TLS markers)
- [x] Verified that Ansible template and config_gen.py explicitly set `host="0.0.0.0"`, so deployed servers are unaffected by the default change
- [ ] Verify on a staging server that Caddy can still reach the router on 127.0.0.1:8080

### DIFF
--- a/compute_space/src/compute_space/config.py
+++ b/compute_space/src/compute_space/config.py
@@ -168,7 +168,7 @@ class DefaultConfig(Config):
     # zone_domain: str
 
     # Server
-    host: str = "0.0.0.0"
+    host: str = "127.0.0.1"
     port: int = 8080
 
     # coredns (only truly needed if tls_enabled)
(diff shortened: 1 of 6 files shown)


zplizzi:
this doesn't work; would break the service interface stuff.

PR #125, opened 2026-05-21
Container networking: hairpin NAT, router bind restriction, ip_forward, network_host

andrewlaack:
Hairpin NAT so containers can reach other apps via public URLs. Pasta shares the host IP but not its listening sockets, so connections to the public IP from inside a container got connection refused. Iptables DNAT rules on the OUTPUT chain rewrite public_ip:80/443 to 127.0.0.1 where Caddy is listening. The boot-time persistence script detects the public IP dynamically so it survives reboots even if the IP changes (EC2 without Elastic IP, etc).

Router bind restricted to 127.0.0.1 + 10.200.0.1 (container gateway) instead of 0.0.0.0. Port 8080 is no longer reachable from the internet. Caddy reaches the router on loopback, containers reach it via the gateway. The gateway bind is auto-detected (skipped in dev/CI where openhost0 doesn't exist).

ip_forward=1 sysctl on the host for containers that need IP forwarding (VPN servers). With --network=host on rootless podman, /proc/sys is read-only so the container can't set this itself.

network_host manifest option (network_host=true in [runtime.container]) runs the container with --network=host instead of pasta. Passes OPENHOST_LOCAL_PORT and sets OPENHOST_ROUTER_URL to 127.0.0.1. Not used by any deployed app currently but available for future use.

### DIFF
--- a/ansible/tasks/containers.yml
+++ b/ansible/tasks/containers.yml
@@ -36,7 +36,7 @@
   when: not _linger_sentinel.stat.exists
   changed_when: true
 
-# net.ipv4.ip_unprivileged_port_start = 25: the router binds rootless on 0.0.0.0:<host_port>, including 25 for SMTP
+# net.ipv4.ip_unprivileged_port_start = 25: apps bind rootless on 0.0.0.0:<host_port>, including 25 for SMTP
 # apps.  Must match compute_space.core.manifest.UNPRIVILEGED_PORT_FLOOR.
 #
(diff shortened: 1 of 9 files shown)


zplizzi (approved):
looks good but yeah i'd like warnings in the docs + in docstrings that network=host disables all network isolation and allows the app to do quite unsafe stuff.

i think there should also be a big warning on the app deploy screen about this.

PR #144, opened 2026-05-28
Add NVIDIA GPU passthrough via CDI

andrewlaack:
When gpu = true in the app manifest, passes --device nvidia.com/gpu=all to podman for CDI-based GPU passthrough.

Changes:
- containers.py: add CDI device flag when manifest.gpu is true
- ansible/tasks/gpu.yml: new task to install NVIDIA drivers and container toolkit, generate CDI spec (skipped on hosts without NVIDIA GPUs)
- setup.yml: include gpu.yml before containers.yml
- test_containers.py: tests for GPU device flag presence/absence

The gpu field already existed in the manifest schema but had no runtime effect. This wires it up.

### DIFF
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -28,6 +28,7 @@
     ansible_python_interpreter: /usr/bin/python3
   tasks:
     - import_tasks: tasks/apt_packages.yml
+    - import_tasks: tasks/gpu.yml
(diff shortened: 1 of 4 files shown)


andrewlaack:
this works, but is (at present) too much complexity to be warranted.

PR #148, opened 2026-06-03
Gate first-boot /setup behind a claim token (CLI + ansible)

kilo-gi:
## Summary
Re-enables the existing claim-token mechanism in `setup_app`, which is already wired up but inert because nothing writes the token file. Without it, anyone who races the operator to `/setup` can set the owner password and take over the instance. This PR adds two opt-in entry points:

- **`openhost up --claim-token <secret>`** — local CLI path. Writes the token to the data dir before launching the router and prints the full claim URL.
- **`ansible-playbook setup.yml ... -e claim_token=<secret>`** — server-deploy path (what vm-manager will use). New task in `setup_openhost_service.yml` writes the token to the same data-dir path before the systemd service starts.

Default is unchanged in both paths: no token supplied → no file written → `/setup` accepts any caller. Safe to merge without coordinating a flip.

Supplied and on-disk tokens are validated as URL-safe (`[A-Za-z0-9_-]+`) in the CLI before embedding in the claim URL; non-URL-safe values exit with an error. Ansible-side validation is left to the caller (vm-manager generates URL-safe tokens; a human passing `-e claim_token=` by hand can debug the bad-URL case themselves).

`setup_app` already deletes the token file on a successful claim, so subsequent boots without `--claim-token` / `-e claim_token` don't resurrect it.

## Test plan
- [x] `pytest self_host_cli/src/self_host_cli/tests/test_cli.py` — new `TestClaimToken` covers: supplied written, supplied overwrites existing, existing preserved when none supplied, no-token default, URL-safety rejection (supplied & on-disk), `token:metadata` parsing.
- [x] Existing `test_claim_token_deleted_after_setup` continues to cover the backend delete-on-claim behavior.
- [x] Manual (ansible, fresh host): `ansible-playbook setup.yml -i <ip>, -e domain=... -e claim_token=abc123` → file present at `/home/host/.openhost/local_compute_space/persistent_data/openhost/claim_token` with mode 0600 before the service comes up; `/setup` enforces the token.
- [x] Manual (ansible, fresh host): same playbook with no `claim_token` extra-var → file absent, `/setup` open (current behavior).

> ⚠️ **Do not test the CLI flag (`openhost up --claim-token`) on an ansible-deployed instance.** `openhost up` rewrites `~/.openhost/local_compute_space/config.toml` from `DefaultConfig`, which collides with the ansible-managed config on the same path and blows away `tls_enabled` / `coredns_enabled` / `start_caddy` / `zone_domain`. Tracked separately as a pre-existing bug. CLI-flag testing should be done on a local dev box, not a deployed VM.

### DIFF
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -9,10 +9,10 @@
 #   -e initial_user=ubuntu        (defaults to root)
 #   -e openhost_commit=<sha>      (git commit to deploy; defaults to main)
 #   -e claim_token=<secret>       (gates /setup until the human supplies this
-#                                  token; required URL-safe value: letters,
-#                                  digits, '-', '_'. Hand the secret to the
-#                                  human who will claim the instance and have
-#                                  them visit /setup?claim=<secret>.)
+#                                  token; URL-safe: letters, digits, '-', '_'.
+#                                  If omitted, the play generates a random
+#                                  token and prints the claim URL. Hand the
+#                                  URL to whoever will claim the instance.)
 
 - name: Bootstrap host user
   hosts: all
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -8,6 +8,11 @@
 #   -e public_ip=1.2.3.4         (defaults to target IP)
 #   -e initial_user=ubuntu        (defaults to root)
 #   -e openhost_commit=<sha>      (git commit to deploy; defaults to main)
+#   -e claim_token=<secret>       (gates /setup until the human supplies this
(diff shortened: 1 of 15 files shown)


zplizzi (approved):
can you add a config variable like `Config.claim_token_required` and use that instead of just checking if the claim file exists in `_claim_token_required`. i want "fail-safe" behavior. and set the default value of `claim_token_required` to True in `DefaultConfig` and anywhere else we setup a Config object (outside of tests).

also double check this gets set when deploying from vm-manager.

otherwise lgtm

PR #153, opened 2026-06-07
Add [tls] cert manifest field for platform cert sharing

andrewlaack:
Apps that do raw TCP TLS (e.g. XMPP s2s federation) can now request the platform wildcard certificate via [tls] cert = true in openhost.toml.

When enabled the platform bind-mounts its cert and key read-only into the container at /run/secrets/tls/tls.crt and /run/secrets/tls/tls.key, and injects OPENHOST_TLS_CERT_PATH / OPENHOST_TLS_KEY_PATH env vars.

The mount is silently skipped when TLS is not configured or cert files don't exist, so apps always start.

Changes:
- manifest.py: add tls_cert field to AppManifest, parse [tls] cert from toml
- containers.py: bind-mount cert/key read-only when tls_cert=True and files exist
- data.py: inject env vars when tls_cert=True and files exist
- apps.py: thread cert paths through deploy and start flows
- tests: manifest parsing, container argv, provision_data env var injection

### DIFF
--- a/compute_space/src/compute_space/core/apps.py
+++ b/compute_space/src/compute_space/core/apps.py
@@ -351,6 +351,8 @@ def insert_and_deploy(
         zone_domain=config.zone_domain,
         port=config.port,
         owner_username=read_owner_username(db) or DEFAULT_OWNER_USERNAME,
+        tls_cert_path=str(config.tls_cert_path) if config.tls_enabled else None,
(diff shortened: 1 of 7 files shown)


andrewlaack:
this is incorrect; this requests the wildcard cert. Closing this, will open a new one with the corrected changes.

PR #160, opened 2026-06-10
tls: per-instance ACME accounts via EAB-mint (default) + byo fallback

kilo-gi:
## What & why

Replaces the **shared** ACME account key (shipped to every VM via ansible at `ansible/secrets/certbot_private_key.json`) with a config-selected `cert_provider`. The shared privileged secret spread across all instances is the problem this removes.

### `eab_mint` (new default, managed)
On first boot each instance creates its **own** ACME account:
1. mints a single-use EAB credential from the central **cert-api** (`POST {cert_api_url}/v1/eab`, per-instance bearer token),
2. generates its own ACME account key locally,
3. registers a new account with **Google Trust Services** using that EAB (RFC 8555 `externalAccountBinding`), against the directory the cert-api returns,
4. persists the account key (`acme-account-key.json` under the openhost data dir).

Renewals reuse the persisted key and **never call the cert-api again**; a lost key just means a fresh EAB next boot. The EAB only authorizes *creating one account* — DNS-01 control still gates issuance.

### `byo` (fallback)
Operator-supplied `acme_account_key_path` + `acme_directory_url`, for non-managed domains. **Fully backward compatible** — this is the pre-EAB-mint behavior.

The DNS-01 issuance core is unchanged and **shared** across both modes; only the "how we get the account key" step differs.

## Changes
- `config.py`: `cert_provider` selector (validated), `cert_api_url`, `cert_api_token`, `managed_acme_account_key_path`.
- `core/tls/cert_api_client.py` *(new)*: `EABCredential` + `mint_eab()` — matches the finalized `openhost-cert-api` `POST /v1/eab` contract (`key_id`/`b64_mac_key`/`directory_url`/`key_algorithm`; `{error,message}` envelope on 401/429/502).
- `core/tls/account.py` *(new)*: `ensure_account_key()` → `ResolvedAccount{account_key, directory_url}` — load persisted key (renewal/byo) or mint+generate+register+persist (eab_mint); account creation and issuance share the cert-api-dictated directory.
- `acquire_cert.py` / `start.py`: thread provider config through; require account key only in `byo`; defense-in-depth `_assert_domains_within_zone` guard (trust boundary = DNS delegation).
- `ansible`: `eab_mint` managed default; shared-key copy now `byo`-only.
- `docs/src/routing.md`: documents both modes, trust boundary, CAA `accounturi` follow-up.
- `scripts/generate_acme_key.py`: reframed as legacy/byo.
- tests: unit coverage (provider selection, `mint_eab` contract + error envelope, zone guard — run in CI) + Pebble EAB end-to-end (`--run-tls`, Linux/manual).

## Contract status (resolved)
The `openhost-cert-api` wire contract is now finalized and this client matches it (`/v1/eab`, bearer auth, `key_id`/`b64_mac_key`/`directory_url`). `cert_api_url` defaults to `https://openhost-cert-api.selfhost.imbue.com` (the minter deployed as the `openhost-cert-api` app).

## Remaining (operational, not code-blocking)
- **Token provisioning**: each instance needs a `cert_api_token` set in its config that also exists in the cert-api's `CERT_API_AUTH_TOKENS` `{label: token}` map. Needs an ansible/provisioning step on both sides.
- **Cutover timing**: ansible default flips to `eab_mint`; hold the live flip until cert-api is deployed. Operators can pin `cert_provider=byo` in inventory during the transition.
- **Renewal directory**: renewals use config's `acme_directory_url` (not re-fetched from cert-api). Fine for the prod-GTS default; if pointing cert-api at staging, set `acme_directory_url` to match (or a future enhancement could persist the directory beside the key).
- **CAA `accounturi` pinning**: follow-up so the CA only honors each instance's own account (instance/DNS side, not a cert-api change).
- Pebble EAB test config keys (`externalAccountBindingRequired`/`externalAccountMACKeys`) to confirm against the Pebble version on first `--run-tls`.

## Testing
- 14 unit tests pass locally (and run in CI). pre-commit (ruff + mypy + secrets) clean. Existing settings/dns/docs tests still pass.
- Pebble EAB end-to-end test is Linux/manual (`--run-tls`); CI does not run it.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -38,6 +38,10 @@ jobs:
       AWS_DEFAULT_REGION: us-east-1
       SKIP_APT_UPGRADE: "1"
       OPENHOST_COMMIT: ${{ github.event.pull_request.head.sha || github.sha }}
+      # Deploy with the bring-your-own ACME account key written below rather than
+      # the default "eab_mint" provider, which would mint an EAB from the
+      # cert-api service (not reachable from the e2e instance).
+      CERT_PROVIDER: byo
       # GCP-specific
       GCP_PROJECT: ${{ secrets.GCP_PROJECT_ID }}
       GCP_ZONE: ${{ vars.GCP_ZONE || 'us-central1-a' }}
--- a/ansible/tasks/sync_code.yml
+++ b/ansible/tasks/sync_code.yml
@@ -16,10 +16,13 @@
   register: git_clean_result
   changed_when: git_clean_result.stdout | length > 0
 
-- name: Copy ACME account key
+# Only needed for bring-your-own cert provider. The default (eab_mint) generates
+# a per-instance ACME account key on the box and never ships a shared key.
+- name: Copy ACME account key (bring-your-own only)
   copy:
     src: "{{ playbook_dir }}/secrets/certbot_private_key.json"
     dest: /home/host/openhost/ansible/secrets/certbot_private_key.json
     owner: host
     group: host
     mode: "0600"
+  when: (cert_provider | default('eab_mint')) == 'byo'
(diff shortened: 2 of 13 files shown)


kilo-gi:
superseded by #163 #164 etc.

PR #162, opened 2026-06-10
Support pre-provisioned API tokens via initial_api_token_hash

andrewlaack:
Lets provisioning tooling such as vm-manager install an owner API token on an instance before first boot. The tooling passes -e initial_api_token_hash=SHA256HEX to ansible, which writes the digest to a file under persistent_data. At boot the router imports the hash into the api_tokens table and deletes the file. The raw token never touches the instance and only works after the owner claims the instance, since the setup-only app has no API routes.

Tested with unit tests and end-to-end on a Hetzner instance provisioned from this branch. Companion PR: imbue-openhost/openhost-vm-manager#29.

### DIFF
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -13,6 +13,13 @@
 #                                  If omitted, the play generates a random
 #                                  token and prints the claim URL. Hand the
 #                                  URL to whoever will claim the instance.)
+#   -e initial_api_token_hash=<sha256-hex>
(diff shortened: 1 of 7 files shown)


andrewlaack:
Closing: not needed. user_id in the provisioning request is just the keycloak user id; no instance-side API token installation is required, and the user-to-instance mapping will live in the signup page app rather than vm-manager.

PR #163, opened 2026-06-10
Add opt-in openhost-cert-api broker as an additional TLS cert provider

kilo-gi:
## Summary

Adds a **second, opt-in TLS cert provider** that fetches certs from the new
`openhost-cert-api` broker (repo: `imbue-openhost/openhost-cert-api`, PR #2),
**alongside** the existing bring-your-own-ACME path. The BYO-ACME path stays the
default and is byte-for-byte unchanged — this is purely additive.

### Why
Today every instance carries a shared ACME credential, so a malicious instance
could mint certs for domains it doesn't control. The broker holds the ACME
account itself and only returns a signed cert for a CSR after the instance proves
DNS control. **Instances never see ACME account credentials.** This PR is the
openhost-side client for that broker.

### Client flow
1. Generate the cert keypair + CSR **locally** — the private key never leaves the
   instance; only the CSR is sent.
2. `POST /v1/orders` with the CSR → receive DNS-01 challenge record(s).
3. Publish the returned `record_name`/`record_value` **verbatim** via the existing
   CoreDNS DNS-01 write path.
4. Poll `POST /v1/orders/{id}/finalize` (202 = keep waiting) with backoff + overall
   timeout; on 200 install the chain + local key via the shared cert-install path.

## Changes
- `core/tls/cert_api_client.py` — sync `httpx` broker client with typed responses
  and 400/401/404 error mapping.
- `core/tls/acquire_cert_broker.py` — the acquisition flow (key/CSR → order →
  publish TXT → poll finalize → install). Deterministic poll loop via an injectable clock.
- `core/dns.py` — `set_txt_records()` writes absolute-FQDN TXT records for broker challenges.
- `config.py` — `cert_provider` (`acme` default / `cert_api`), `cert_api_base_url`
  (defaults to `https://api.selfhost.imbue.com`), `cert_api_token`. Backward compatible.
- `web/start.py` — selects provider; ACME path unchanged.
- `ansible/templates/config.toml.j2` — opt-in `cert_api` block.
- Shared `write_cert_and_key` / `tls_private_key_to_pem` reused by both providers.

## Testing
TDD against a **faked broker** (`httpx.MockTransport`) — no real service is contacted:
- `test_cert_api_client.py` — request shapes, auth header, JSON parsing, error mapping (9).
- `test_acquire_cert_broker.py` — full flow: CSR-only (no private key) leaves, TXT
  published before polling, 202→200 polling, cert+key install with 0600 perms, TXT
  cleanup on success **and** timeout (3).
- `test_dns.py` — absolute-FQDN TXT writes + cleanup (3 new).
- `test_cert_provider_config.py` — default stays `acme`, legacy configs load, round-trip (4).

All green locally; ruff + mypy (strict) clean.

## Notes / open items
- **Broker URL:** `https://api.selfhost.imbue.com` is wired as the default but the
  service is **not yet deployed** — a DNS record will point there when it goes up.
- Broker flow is **synchronous** (called directly from `start.py`), unlike the BYO
  path's `asyncio.run(...)` — simpler and easier to test deterministically.
- Requests a **wildcard CSR** (`[domain, *.domain]`), matching today's BYO behavior
  (confirmed the broker accepts multi-SAN/wildcard CSRs).

### DIFF
--- a/ansible/templates/config.toml.j2
+++ b/ansible/templates/config.toml.j2
@@ -14,3 +14,14 @@ start_caddy = true
 claim_token_required = true
 data_root_dir = "/home/host/.openhost/local_compute_space"
 apps_dir_override = "/home/host/openhost/apps"
+{% if cert_provider is defined and cert_provider == 'cert_api' %}
+# Opt-in: fetch TLS certs from the openhost-cert-api broker instead of the
+# bring-your-own ACME account key above (which is then ignored).  The broker
+# holds the ACME account, so this instance never sees ACME credentials.
+# cert_api_base_url defaults to the canonical broker; only set it to override.
+cert_provider = "cert_api"
+{% if cert_api_base_url is defined %}
+cert_api_base_url = "{{ cert_api_base_url }}"
+{% endif %}
+cert_api_token = "{{ cert_api_token }}"
+{% endif %}
--- a/compute_space/src/compute_space/config.py
+++ b/compute_space/src/compute_space/config.py
@@ -9,6 +9,13 @@ import cattrs
 import tomli_w
 import typed_settings
 
+# TLS cert provider selection (see Config.cert_provider).
+# "acme" is the default bring-your-own-ACME-credentials path (unchanged, fully
+# backward compatible). "cert_api" fetches certs from the openhost-cert-api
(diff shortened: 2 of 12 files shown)


zplizzi (approved):
codeowner-covered changes look fine

PR #164, opened 2026-06-11
cert_api: authenticate the broker with Keycloak client-credentials

kilo-gi:
## What

Instance-side half of the cert-api **Keycloak auth cut-over**. Swaps the broker's static per-instance bearer token (`cert_api_token`, added in the base branch) for a per-instance **Keycloak confidential client** (service account) in the `openhost-customers` realm.

The instance fetches a short-lived access token from its issuer via the OAuth2 **client-credentials** grant and presents it to cert-api as a bearer. cert-api validates the JWT and enforces that the CSR's SANs fall within the instance's assigned subdomain. No shared secret or ACME account key is needed on the `cert_api` path.

Stacked on `kilo/sculptor/cert-api-client` and targets it (not `main`).

## Changes

- **`core/tls/keycloak.py`** (new): `KeycloakClientCredentials` + a `TokenProvider` protocol with `StaticTokenProvider` (tests / BYO) and `KeycloakTokenProvider`. The Keycloak provider POSTs `grant_type=client_credentials` to `<issuer>/protocol/openid-connect/token` and **caches the token to ~expiry** — access tokens last ~300s while the finalize-poll loop can run minutes, so it refetches ~30s before expiry.
- **`core/tls/cert_api_client.py`**: `CertApiClient` now holds a `TokenProvider` and builds the `Authorization` bearer header **per request**, so the token can refresh transparently mid-flow during polling. `/health` stays unauthenticated.
- **`config.py`**: drop `cert_api_token`; add `cert_api_keycloak_issuer_url`, `cert_api_keycloak_client_id`, `cert_api_keycloak_client_secret` (all injected by provisioning; `client_secret` is the only sensitive one — treated like the ACME account key). `start.py` builds the provider and **fails loudly** if any of the four cert_api settings are unset.
- **`ansible/templates/config.toml.j2`**: render the three Keycloak fields under the `cert_api` block in place of `cert_api_token`.

## Reused as-is (from the base branch)

The CSR build (`[domain, *.domain]`, private key never leaves the instance), the CoreDNS DNS-01 TXT write path (`set_txt_records` / verbatim broker challenge values), and the finalize-poll loop are unchanged — this PR only changes **how the broker is authenticated**.

## Scope note (please confirm)

The original task framing said to "remove the direct-ACME path". I **kept the `cert_api` provider opt-in alongside the default BYO-ACME path**, because the base branch is deliberately additive (its tests assert ACME stays the default and legacy configs keep loading), the broker is still only at a QA URL (not production-deployed), and removing ACME now would break cert acquisition on all existing instances. For the `cert_api` path itself, no ACME account key is loaded — which matches the cut-over intent. Happy to do the full ACME removal as a follow-up once the broker is production-deployed and instances are migrated, if that's what you want here.

## Testing

- New `tests/test_keycloak.py`: client-credentials POST shape, token endpoint composition, cache-until-near-expiry + refresh, error/`missing access_token` handling (MockTransport + fake monotonic clock).
- Updated `test_cert_api_client.py` (incl. a new test proving the bearer is fetched fresh per request), `test_acquire_cert_broker.py`, `test_cert_provider_config.py` for the new config/token-provider shape.
- Full lightweight suite green: **667 passed, 35 skipped**. `ruff` + `mypy --strict` + pre-commit clean.

### DIFF
--- a/compute_space/src/compute_space/web/start.py
+++ b/compute_space/src/compute_space/web/start.py
@@ -26,6 +26,8 @@ from compute_space.core.tls.acquire_cert import acquire_tls_cert
 from compute_space.core.tls.acquire_cert import check_if_cert_exists
 from compute_space.core.tls.acquire_cert_broker import acquire_tls_cert_via_broker
 from compute_space.core.tls.cert_api_client import CertApiClient
+from compute_space.core.tls.keycloak import KeycloakClientCredentials
+from compute_space.core.tls.keycloak import KeycloakTokenProvider
 from compute_space.core.updates import RESTART_EXIT_CODE
 from compute_space.core.updates import initialize_shutdown_event
 from compute_space.db import init_db
@@ -62,6 +64,13 @@ def _owner_exists(config: Config) -> bool:
         db.close()
 
 
+def _require_cert_api_setting(value: str | None, name: str) -> str:
+    """Return a required cert_api config value, failing loudly if unset."""
+    if not value:
+        raise RuntimeError(f"{name} must be set in config to use the cert_api provider")
+    return value
+
+
 def _acquire_missing_tls_cert(config: Config) -> None:
     """Acquire a missing TLS cert using the configured provider.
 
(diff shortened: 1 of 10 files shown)


compute_space/src/compute_space/web/start.py
```diff
@@ -84,18 +93,25 @@ def _acquire_missing_tls_cert(config: Config) -> None:
...
-            raise RuntimeError("cert_api_token must be set in config to use the cert_api provider")
-        with CertApiClient.create(config.cert_api_base_url, config.cert_api_token) as client:
-            acquire_tls_cert_via_broker(
-                domain=config.zone_domain,
-                cert_path=config.tls_cert_path,
-                key_path=config.tls_key_path,
-                coredns_zonefile_path=config.coredns_zonefile_path,
-                client=client,
-            )
+        base_url = _require_cert_api_setting(config.cert_api_base_url, "cert_api_base_url")
```
zplizzi, line 96: nit - i would put this validation logic in Config, probably as a `__attrs_post_init__` so it can apply only when `cert_provider == CERT_PROVIDER_CERT_API`. it's nicer to centralize validation all in once place, and then all users can trust that any Config object is valid.

zplizzi (approved):
i just looked at the start.py changes (code-owner-covered), those look fine to me!

PR #167, opened 2026-06-15
Centralize cert_api config validation + DNS/TOML cleanups

kilo-gi:
## Summary

A series of cert_api / DNS / config cleanups on top of centralizing cert_api config validation.

- **Centralize cert_api config validation** in `Config.__attrs_post_init__` so invalid cert_api configuration is caught at construction time.
- **Use absolute imports** instead of relative in the compute_space tests.
- **cert_api: harden key file perms and bound the finalize poll loop** — create the TLS private key with `0o600` from the start (`os.open`) so it is never briefly world-readable between write and chmod; restructure `_poll_until_issued` to loop while within the deadline so the timeout is a single, clear exit condition.
- **dns: collapse `set_txt`/`set_txt_records` into one `append_txt_records`** — the two functions shared ~90% of their body and differed only in name addressing (relative-to-`$ORIGIN` vs absolute FQDN). The unified function writes each `TxtRecord.record_name` verbatim and lets the caller choose addressing; the broker path normalizes its FQDNs via a new `_as_fqdn` helper.
- **config: share TOML dict-building between `to_toml` and `to_toml_str`** — extract `_to_toml_dict` so both writers share one source of truth.

## Testing

- `pixi run -e dev pytest -x compute_space/src/compute_space/tests/test_dns.py compute_space/src/compute_space/tests/test_acquire_cert_broker.py` — 9 passed
- `pixi run -e dev pytest -x compute_space/src/compute_space/tests/ -k "toml or config"` — 22 passed
- pre-commit (ruff + mypy) clean on all commits

### DIFF
--- a/compute_space/src/compute_space/web/start.py
+++ b/compute_space/src/compute_space/web/start.py
@@ -12,7 +12,6 @@ import hypercorn.asyncio
 import hypercorn.config
 
 from compute_space.config import CERT_PROVIDER_ACME
-from compute_space.config import CERT_PROVIDER_CERT_API
 from compute_space.config import Config
 from compute_space.config import load_config
 from compute_space.config import set_active_config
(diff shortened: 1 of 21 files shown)


compute_space/src/compute_space/web/start.py
```diff
@@ -64,19 +63,16 @@ def _owner_exists(config: Config) -> bool:
...
-
-
 def _acquire_missing_tls_cert(config: Config) -> None:
     """Acquire a missing TLS cert using the configured provider.
 
     Caller has already verified the cert is missing, CoreDNS is enabled, and
     acquire_tls_cert_if_missing is set.  The default "acme" provider is the
     unchanged BYO-ACME path; "cert_api" fetches from the openhost-cert-api broker.
+
+    The cert_provider value and its required settings are validated when the
```
zplizzi, line 73: nit: redundant comment, the one below feels appropriate and the one in the docstring feels like claude comment slop

zplizzi (approved):
LGTM, def appreciate the minor cleanups you did along the way :)

PR #179, opened 2026-06-18
Fix X-Forwarded-Proto/For not reaching apps correctly behind Caddy

zplizzi:
## Problem

Caddy terminates TLS and proxies to hypercorn over plain HTTP on loopback. The subdomain proxy stripped the inbound `X-Forwarded-*` headers and rebuilt them from that internal hop, discarding the authentic values Caddy had set:

- **`X-Forwarded-Proto`** was set from `scope["scheme"]`, which is *always* `"http"` behind Caddy. So **every app behind TLS saw `X-Forwarded-Proto: http`** — breaking secure-cookie flags, absolute-URL/redirect building, OAuth `redirect_uri`, HSTS, and framework "am I on HTTPS?" checks. (The codebase already worked around this exact issue in `build_login_url`.)
- **`X-Forwarded-For`** was set from `connection.client.host`, which behind Caddy is *always* `127.0.0.1`. Apps never saw the real client IP (rate-limiting, geo, audit logs).
- The **WebSocket** path set no `X-Forwarded-Proto` at all.

Caddy's documented default (no `trusted_proxies` is configured) is to strip client-spoofed `X-Forwarded-*` and set authentic values — so the values reaching hypercorn from Caddy are trustworthy, and the proxy was throwing them away.

## Fix (`subdomain_proxy.py`)

- **`X-Forwarded-Proto`** → derived from `config.tls_enabled` (`https`/`http`), now set for both HTTP and WebSocket. Not spoofable; correct because the `:80→:443` redirect means nothing is ever proxied in the clear when TLS is on.
- **`X-Forwarded-For`** → new `_resolve_forwarded_for()`: trusts the inbound XFF **only when the peer is loopback** (Caddy), otherwise falls back to the peer address — so a container on the `10.200.0.x` gateway can't spoof it.
- **`X-Forwarded-Host`** → unchanged (already used the original Host).

## Tests

- New `tests/test_subdomain_proxy.py` unit tests for `_resolve_forwarded_for` (loopback trusts inbound, IPv6 loopback, **non-loopback ignores spoof**, fallback, no-client) — runs in the default suite.
- Updated the integration anti-spoof test to the trust-front-proxy model and fixed it to normalize header case (the old assertions were silently vacuous).
- Strengthened the e2e test to assert `x-forwarded-proto == "https"` end-to-end through real Caddy.

`609 passed, 32 skipped` (container/TLS skips on macOS) + 5 new unit tests. `mypy`/`ruff` clean.

### DIFF
--- a/compute_space/src/compute_space/tests/test_integration.py
+++ b/compute_space/src/compute_space/tests/test_integration.py
@@ -1103,23 +1103,28 @@ class TestContainerE2E:
         assert headers_ci.get("x-custom-test") == "test-value"
         assert "x-forwarded-for" in headers_ci
         assert "x-forwarded-host" in headers_ci
+        assert headers_ci.get("x-forwarded-proto") == "http"  # tls_enabled is False in tests
 
-    def test_proxy_strips_spoofed_forwarded_headers(self, admin_session, config):
-        """Client-supplied X-Forwarded-* headers are overwritten, not forwarded."""
+    def test_proxy_forwarded_headers_trust_model(self, admin_session, config):
+        """X-Forwarded-Proto/Host are derived by the router, never taken from the
+        client.  X-Forwarded-For is trusted only from the loopback front proxy —
+        which the test client is — so its value is passed through to the app."""
         r = admin_session.get(
             f"{_app_url(config, 'test-app')}/echo-headers",
             headers={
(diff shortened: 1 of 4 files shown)


zplizzi:
cc @kilo-gi

PR #187, opened 2026-06-19
Add support for running a server on a Raspberry Pi 5. 

samuelren-imbue:
Changes for RPi compatibility:
- add aarch64 to the supported platforms list
- download coredns from git directly, since aarch64 distribution not available through conda
- enable cgroup (disabled on rpi by default)
- some apparmor configuration? i dont really understand this part

Other changes:
- added support for lan testing with an option to disable tls (also dont really get this part)

Issue: 
- currently all the apps are built only for linux x86 or mac arm64, so the rpi cant run it at the moment

The current way to run it locally is:
1. Flash an SD card with Raspberry Pi OS Lite 64 bit
2. SSH into it with `ssh -L 8080:localhost:80 <hostname>`. This is to open the webpages later on localhost
3. Run: `curl -fsSL https://raw.githubusercontent.com/imbue-openhost/openhost/refs/heads/rpi-support/scripts/provision.sh | bash -s -- --domain localhost:8080 --no-tls --branch rpi-support` with root
(it will restart at this stage, will need to reconnect)
4. Get the setup token from `sudo cat /home/host/.openhost/local_compute_space/persistent_data/openhost/claim_token`
5. Go to setup page at: https://localhost:8080/setup?claim=<token> and proceed normally.


ansible/tasks/apt_packages.yml
```diff
@@ -30,6 +30,36 @@
       - fuse3
     state: present
 
+# coredns is not published for linux-aarch64 on conda-forge, and is not in the
```
zplizzi, line 33: let's install from github releases for all platforms - better to be consistent

ansible/tasks/containers.yml
```diff
@@ -55,7 +55,9 @@
     mode: "0644"
 
 - name: Apply podman sysctls
-  command: sysctl -p /etc/sysctl.d/90-openhost-podman.conf
+  # --ignore-unknown skips keys that don't exist on this kernel (e.g.
+  # kernel.apparmor_restrict_unprivileged_userns on kernels without AppArmor).
+  command: sysctl --ignore-unknown -p /etc/sysctl.d/90-openhost-podman.conf
```
zplizzi, line 60: i'd prefer to instead not write flags into the config than to ignore them - the latter allows bugs in the config to get silenced/ignored. but maybe we don't need this change at all if we switch to ubuntu rpi image.

ansible/tasks/apt_packages.yml
zplizzi: and should simplify a lot of this PR

ansible/tasks/pixi.yml
```diff
@@ -39,27 +53,34 @@
 - name: Set cap_net_bind_service on coredns
   become: yes
   become_user: root
-  shell: setcap 'cap_net_bind_service=+ep' {{ coredns_path.stdout }}
+  shell: setcap 'cap_net_bind_service=+ep' {{ coredns_bin | trim }}
 
-- name: Create resolved.conf.d directory
+- name: Check if systemd-resolved is active
```
zplizzi, line 58: switching to ubuntu base will hopefully mean we don't need these changes

scripts/provision.sh
```diff
@@ -93,6 +96,42 @@ else
 fi
 chown -R host:host "$OPENHOST_DIR"
 
+# ---- Enable cgroup memory controller if needed (Raspberry Pi) ----
```
zplizzi, line 99: switching to ubuntu will hopefully remove the need for this

zplizzi (commented):
i think you should use an ubuntu base for the pi image, instead of the custom raspberry pi OS. that'll keep things more consistent between our different installs.

PR #197, opened 2026-06-29
Add logout button to dashboard nav

andrewlaack:
Surfaces the existing POST /logout endpoint in the shared dashboard nav as a Log out button, alongside a Logged in as <name> indicator. Previously there was no way to log out from the UI; the backend route already existed but was never wired to anything.

The button is a top-level form POST to /logout (the session cookie is httponly + samesite=lax, so logout must round-trip through the server). It lives in the shared layout.html, so it appears on every authenticated page (dashboard, settings, system, terminal, app detail, add app) and is correctly absent from the standalone login page.

This PR also fixes a pre-existing forced-logout CSRF on POST /logout. Because /logout has no owner-auth guard (it must work for any session state), it was the one state-changing POST that skipped the Origin check every other endpoint gets via verify_owner_auth, so a cross-site page could force-log-out the operator. Added a verify_same_origin helper that rejects any request whose Origin header is present but does not parse to the exact target host (this includes Origin: null sent by sandboxed iframes), and applied it to /logout. Same-origin requests (matching Origin, or no Origin on a top-level form post) still log out normally.

Tests: render test for the button; route tests that POST /logout revokes the session and clears the cookie, is no-op-safe with no session, rejects cross-origin/null/lookalike/wrong-port origins (and leaves the victim session intact), and allows same-origin and no-Origin requests.

Verified live on Hetzner instances provisioned from this branch via vm-manager: the real browser logout button works and revokes the session; cross-origin, Origin: null, and wrong-port POSTs do not revoke the session; same-origin and no-Origin requests do. Also ran extensive manual and adversarial edge-case suites (cookie/token fuzzing, HTTP method fuzzing, open-redirect checks, multi-tab logout, JS-disabled logout, etc) across multiple fresh instances.

### DIFF
--- a/compute_space/src/compute_space/web/auth/auth.py
+++ b/compute_space/src/compute_space/web/auth/auth.py
@@ -38,7 +38,13 @@ def _get_bearer_token_if_set(connection: AnyConnection) -> str | None:
 def get_connection_origin(connection: AnyConnection) -> str | None:
     """gets and formats the origin header as "sub.example.com" or "sub.example.com:1234", no protocol or path, if set.
     port is included if non-default.
-    returns None if no origin header is set.
+    returns None only if no Origin header is set at all.
+
+    A *present* Origin that has no parseable host (notably ``Origin: null``, which browsers send from
+    sandboxed/opaque contexts like ``<iframe sandbox>``) returns the literal token (e.g. ``"null"``)
+    rather than None, so callers can't mistake a present cross-origin request for an absent header and
+    wave it through. Such a token never equals a real ``host[:port]``, so origin-match checks fail
+    closed.
 
     browsers have specific behaviors around the Origin header, which we rely on here.
     - it is set on all cross-origin requests, including from subdomains.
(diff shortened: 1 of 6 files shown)


compute_space/src/compute_space/web/auth/auth.py
```diff
@@ -56,6 +56,30 @@ def get_connection_origin(connection: AnyConnection) -> str | None:
...
+    request is cross-site and is rejected. This stops a hostile page (including a sandboxed iframe
+    sending ``Origin: null``) from cross-site POSTing to endpoints like /logout, which has no
+    owner-auth guard of its own (it must work for any session state).
+
+    A genuinely same-origin top-level form post either omits Origin or sends the matching host, so
+    legitimate logout still works.
+
+    raises NotAuthorizedException on a cross-origin request.
+    """
+    # Use the raw header (not get_connection_origin) so that a present-but-unparseable Origin such
```
zplizzi: seems like maybe this logic should go in get_connection_origin? not sure

compute_space/src/compute_space/web/routes/pages/login.py
```diff
@@ -64,6 +65,9 @@ async def login_post(request: Request[Any, Any, Any], db: sqlite3.Connection, co
 
 @post("/logout", status_code=200)
 async def logout(request: Request[Any, Any, Any], db: sqlite3.Connection, config: Config) -> Response[Any]:
+    # /logout has no owner-auth guard (it must work for any session state), so guard it against
+    # cross-site POSTs here to prevent forced-logout CSRF.
+    verify_same_origin(request)
```
zplizzi: nit: litestar has "guards" which maybe is the more correct pattern for this. all good if you prefer this tho

zplizzi (approved):
lgtm!

compute_space/src/compute_space/web/auth/auth.py
andrewlaack: Good call — moved the opaque-origin handling into get_connection_origin: it now returns a non-None token (e.g. "null") for a present-but-hostless Origin instead of collapsing to None, so every origin-match caller (verify_owner_auth/verify_app_auth/verify_same_origin) fails closed on Origin: null rather than treating it as "no header". verify_same_origin is back to the simple origin == target-netloc check. Done in 311621b.

compute_space/src/compute_space/web/routes/pages/login.py
andrewlaack: Switched to a litestar guard — added require_same_origin and attached it via @post("/logout", guards=[require_same_origin]), matching the require_owner_auth pattern. Done in 311621b.

PR #199, opened 2026-06-29
Allow ssh into app

samuelren-imbue:
Allow ssh into app via `oh app ssh cool-app`, runs /bin/sh in the container. In the future, it could use an external binary, but would require other setup.


andrewlaack:
1) The remote command interpolates `container_id` and `--shell` into a shell string without quoting (possibly dangerous)
2) add `oh app ssh` to the command list in the CLI README

andrewlaack (approved):

PR #201, opened 2026-06-30
apps show "starting" rather than "stopped" on init

samuelren-imbue:
bug fix for initial space setup.

### DIFF
--- a/compute_space/src/compute_space/core/startup.py
+++ b/compute_space/src/compute_space/core/startup.py
@@ -1,12 +1,18 @@
 import os
 import sqlite3
 import threading
+import time
 
 from compute_space.config import Config
 from compute_space.core.apps import start_app_process
 from compute_space.core.containers import is_container_running
 from compute_space.core.default_apps import deploy_default_apps
 from compute_space.core.logging import logger
+
+# UTC timestamp captured at module import.  check_app_status uses this to
+# distinguish rows inserted by this process (whose build threads are still
+# running) from rows abandoned by a previous process (which should be swept).
+_PROCESS_START_UTC = time.strftime("%Y-%m-%d %H:%M:%S", time.gmtime())
 
 
 def check_app_status(config: Config) -> None:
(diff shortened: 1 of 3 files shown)


compute_space/src/compute_space/core/startup.py
```diff
@@ -56,10 +56,23 @@ def check_app_status(config: Config) -> None:
...
+                            row["app_id"],
+                        ),
+                    )
+                    continue
                 db.execute(
-                    "UPDATE apps SET status = 'stopped' WHERE app_id = ?",
+                    "UPDATE apps SET status = 'starting' WHERE app_id = ?",
                     (row["app_id"],),
                 )
+                apps_to_restart.append(row["app_id"])
```
andrewlaack: First-boot race: `/setup` runs `deploy_default_apps`, which inserts apps as `building`/`container_id=NULL` with a background build thread then the same process hits `check_app_status` while those builds are still running. This branch treats them as interrupted and schedules a *second* concurrent build, the exact hazard the sequential restart thread exists to avoid (racing `podman rm -f`, last-writer-wins on `container_id`, token regen mismatch). Can we skip in-flight rows, e.g. a boot-id/PID stamp or only sweeping `container_id IS NULL` rows older than process start?

compute_space/src/compute_space/core/startup.py
```diff
@@ -56,10 +56,23 @@ def check_app_status(config: Config) -> None:
                 )
                 apps_to_restart.append(row["app_id"])
             else:
```
andrewlaack: This block is the same as the `if row["container_id"]:` branch above, and `alive` already covers the no-container case

compute_space/src/compute_space/tests/test_check_app_status.py
```diff
@@ -129,6 +129,42 @@ def test_running_app_with_dead_container_is_restarted(tmp_path: Path, monkeypatc
     assert app_id in restarted
```
andrewlaack, line 139: can you write a test for the new error guard? A no-container app with a missing/invalid `repo_path` should be marked `error` and not restarted

andrewlaack (changes requested):

andrewlaack (approved):
LGTM

PR #202, opened 2026-07-01
Standardize dash/underscore in openhost_system_agent

samuelren-imbue:
The CLI and HTTP requests now use dash. These are internal tools and this update is atomic, so the server will be fine. If any apps hit these, then they will need to update.

Python package maintains underscore since it can't parse dashes.


andrewlaack (approved):
LGTM

PR #204, opened 2026-07-01
Dashboard UI Update

samuelren-imbue:
Updated the main dasboard UI
1. nav bar link color doesn't change by link, and doesn't always have underline
2. details link replaced by button buttons vs links
3. information ordering in table by importance
4. show tabs along the top rather than links

### DIFF
--- a/compute_space/src/compute_space/web/templates/layout.html
+++ b/compute_space/src/compute_space/web/templates/layout.html
@@ -14,7 +14,7 @@
     button, input[type="submit"], .btn { padding: 0.4em 1em; cursor: pointer; border: 1px solid #ccc; background: #f9f9f9; text-decoration: none; color: #222; display: inline-block; font: inherit; font-size: 0.875em; }
     .btn-danger { background: #fee; border-color: #c00; color: #c00; }
     .btn-primary { background: #e8f0fe; border-color: #36c; color: #36c; }
-    nav { display: flex; align-items: flex-end; gap: 0.25em; }
+    nav { display: flex; align-items: flex-end; gap: 0.25em; border-bottom: 1px solid #e8e8e8; }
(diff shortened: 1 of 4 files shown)


compute_space/src/compute_space/web/templates/layout.html
```diff
@@ -42,20 +47,30 @@
...
       <button type="submit" class="btn btn-danger btn-logout">Log out</button>
     </form>
   </nav>
-  <hr>
+  <script>
+    (function() {
+      var p = window.location.pathname;
+      document.querySelectorAll('#main-nav .nav-tab').forEach(function(a) {
+        var href = a.getAttribute('href');
+        if (href !== '/docs/' && (p === href || (href !== '/dashboard' && p.startsWith(href)))) {
```
andrewlaack: The active-tab highlight won't fire on /. The dashboard serves both / and /dashboard, and login redirects to /, but the script excludes /dashboard from prefix matching, so no tab is active on the default landing page. Also /terminal won't match href /terminal/. Maybe special-case / to Dashboard and handle the trailing slash.

compute_space/src/compute_space/web/templates/terminal.html
```diff
@@ -1,9 +1,6 @@
 {% extends "layout.html" %}
 {% block title_suffix %} - Terminal{% endblock %}
 {% block content %}
```
andrewlaack, line 3: Dropping body max-width none from terminal.html clamps the terminal to 960px (set in layout.html). logs.html still keeps this override, so it looks load-bearing, probably want to keep it here too.

andrewlaack (changes requested):

andrewlaack (approved):

PR #205, opened 2026-07-02
Log cleanup

samuelren-imbue:
Add rotating logs
Remove app logs from going to system journald
Archive up to 5 app startup and runtime logs
Add documentation for which logs go where


andrewlaack (approved):
LGTM

PR #207, opened 2026-07-07
Move tokens menu into settings

samuelren-imbue:
Also moved inline js from settings.html and add_app.html to settings.js and add-app.js.


andrewlaack (approved):
LGTM

PR #211, opened 2026-07-08
Proxy: don't cut off long-poll requests (fixes Matrix /sync 504 disconnections)

andrewlaack:
Fixes intermittent Matrix client disconnections caused by the router cutting off long-poll requests.

Problem
The router proxies app HTTP requests with a single 30s httpx timeout (proxy_http_request, timeout=30), which also applies to the read timeout. Matrix clients keep their session alive by long-polling GET /_matrix/client/v3/sync for ~30s waiting for events. When a sync holds the full window with no events, the 30s read timeout fires before the backend responds and the router returns 504 "App timed out". matrix-js-sdk treats the 504 as a sync error and surfaces a disconnection to the user. It's intermittent because it only triggers when a sync happens to hold ~30s idle (busy rooms return sooner and dodge it).

Root cause is proxy.py: httpx.AsyncClient(timeout=30) sets connect/read/write/pool all to 30s, and subdomain_proxy calls it without overriding the timeout.

Fix
Keep a short connect timeout (fail fast for a dead/not-listening app -> gateway error) but leave the read timeout open so long-poll and streaming endpoints (Matrix /sync, SSE, hanging GETs) aren't cut short. A bare-number or None timeout now shapes only connect/write/pool; callers can pass an httpx.Timeout for full control. This mirrors how the WebSocket proxy already treats long-lived connections (no read timeout).

Testing
Unit tests for the timeout shaping, plus integration tests against a real slow backend: a long-poll that outlasts a tiny connect timeout still returns 200, a dead app fails fast with a gateway error, normal fast requests are unchanged, and an opt-in test (RUN_SLOW_PROXY_TEST=1) proves a 31s response is no longer cut to 504.

Also verified end-to-end on a real Hetzner OpenHost instance running this branch: deployed openhost-synapse, and 8 idle /sync long-polls all returned 200 holding the full ~50s with zero 504s (on stock they 504 ~half the time). Fast requests (~1s) and the app self-restart path (~3s) were unaffected.

ruff and mypy clean; existing proxy/provider test suites pass. The 12 failing tests in test_integration.py are pre-existing environmental failures (test router can't bind in this sandbox) and fail identically on main without this change.

### DIFF
--- a/compute_space/src/compute_space/tests/test_proxy_timeout.py
+++ b/compute_space/src/compute_space/tests/test_proxy_timeout.py
@@ -1,0 +1,217 @@
+"""Tests for the proxy's timeout behavior.
+
+Regression coverage for the intermittent ``504 App timed out`` that broke
+Matrix ``/sync`` long-polls: the router proxied app requests with a single 30s
+timeout, so a long-poll that held ~30s with no events was cut off and the
+client saw a disconnection.  The fix keeps a short *connect* timeout (fail fast
+for dead apps) but a generous *read* timeout so long-polls aren't killed.
+"""
+
+from __future__ import annotations
+
+import os
+import socket
+import threading
+import time
+from http.server import BaseHTTPRequestHandler
+from http.server import ThreadingHTTPServer
+from typing import Any
+
+import httpx
+import pytest
+from litestar import Request
+from litestar.response.base import ASGIResponse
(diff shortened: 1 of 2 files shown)


andrewlaack:
Closing in favor of an application-side fix (limiting the Matrix sync long-poll below the router's proxy timeout) so the change lives in the app rather than the platform.

PR #225, opened 2026-07-13
arm support

samuelren-imbue:
add support for arm running ubuntu for raspberry pi project


ansible/tasks/coredns.yml
```diff
@@ -0,0 +1,67 @@
...
+
+- name: Pin CoreDNS release
+  set_fact:
+    coredns_version: "1.14.4"
+    coredns_arch: "{{ 'arm64' if ansible_architecture in ['aarch64', 'arm64'] else 'amd64' }}"
+    coredns_sha256:
+      amd64: "5b0e6a6a8b97bdcaec65baa70e83ca88ce03ab852355c656e3f7953405cfe36e"
+      arm64: "b3491ed0fbe530a549640a4073d90037ffafa483f007936e1943c1ae0def7325"
+
+# ---- amd64: CoreDNS was installed by pixi (see pixi.yml) ----
```
zplizzi: let's install from github for all arches - better to be consistent.

compute_space/src/compute_space/core/pinned_binaries.toml
```diff
@@ -0,0 +1,19 @@
+# Arch-specific downloads for binaries compute_space installs at RUNTIME.
```
zplizzi: i'd probably just put this into pinned_binary.py as a python object instead of as a toml? not sure what the toml is getting us.

zplizzi: and then you save the parsing code etc

zplizzi (approved):
lgtm with minor feedback!

PR #227, opened 2026-07-15
fix coredns issue on upgrade

samuelren-imbue:
check coredns installation at runtime
works both from restarting a previous broken instance and working old instance


zplizzi (approved):

zplizzi:
@kilo-gi no need to review, this LGTM

zplizzi:
btw i also requested to get added to the maintainer list for the conda package, so maybe we can switch back to pixi if that goes through and i get it updated

PR #233, opened 2026-07-17
Archive tier: always JuiceFS (local file-backed by default), JuiceFS-native local->S3 migration

andrewlaack:
Makes the app_archive storage tier always available so apps that request it install on any zone, and lets an operator later upgrade a zone to durable S3 with an in-product data migration.

Reworked per review (Zack): instead of forking the app-mount logic between a plain local directory and a JuiceFS-on-S3 mount, the archive tier is now ALWAYS a JuiceFS volume mounted at the same host path. Only JuiceFS's object storage differs by backend:

- local (the default): JuiceFS --storage file, objects live on local disk under persistent_data/app_archive_local_objects (backed up). Formatted + mounted at boot by attach_on_startup, so the tier is available with zero operator config.
- s3: JuiceFS --storage s3 against an operator-supplied bucket.

Because it's always the same mount, app containers always bind-mount the JuiceFS mountpoint; effective_archive_dir always returns it. All the local-vs-s3 branching, the bespoke per-file copy under podman unshare, ownership reproduction, and the short-copy verifier are gone.

Migration local -> s3 is now JuiceFS-native and provider-agnostic: quiesce archive-using apps (so no writes race the copy and nothing holds the mount open), juicefs sync the underlying objects into the bucket (--check-all verified; --no-https for http endpoints; AWS_* env creds), juicefs config re-points the same volume at S3 (metadata untouched, so every file/dir/mode/owner is preserved), remount, flip the DB row to s3, restart the apps, then reclaim the local object store. The same sync+config mechanism generalizes to migrating between two S3 providers. Fail-open: any failure before the DB flip leaves the volume reading from the intact local store and the backend at local; a state_message tells the operator to restart if the local remount couldn't complete.

Guarded behind an explicit confirm_migrate_local acknowledgement; the dashboard lists which apps' data will migrate. Install-time storage summary + durability notice retained. Docs (data.md, manifest_spec.md, creating_an_app.md), dashboard copy, and the systemd unit (juicefs umount -f, TimeoutStopSec) updated.

Tests: unit + API tests rewritten for the new model, plus a 61-case edge-case module; the e2e suite now seeds local archive data, asserts the confirm gate, migrates to a local MinIO, and verifies the pre-migration data survives byte-identical.

Verified end-to-end on live Hetzner instances (fresh, committed code, ansible-installed unit): local file-backed JuiceFS comes up at boot with zero config; full local -> S3 migration with archive apps running; pre-migration data survives byte-identical; post-migration writes flush to the bucket; local object store reclaimed. vet (opus + fable agentic, with and without history) clean after addressing findings; CI green (pre-commit, test, test-containers, e2e gcp + ec2).


andrewlaack:
received dm from zack about this. will update it to always use juicefs just the local one is a local s3 store.

PR #236, opened 2026-07-19
Enable the storage guard by default

andrewlaack:
The storage guard stops running apps when free disk drops below a threshold, so a full disk doesn't silently take an instance fully down (generic 500s). It was disabled by default (storage_min_free_mb defaulted to 0), so owners got no protection unless they knew to set the config key and reboot — which is what bit us in practice.

This simply enables the guard by default: storage_min_free_mb now defaults to 500 (MB) instead of 0. The guard stays config-only — operators change the threshold, or disable it entirely with 0, in the router config and reboot. No new UI, table, migration, or API surface (keeping the configurable-setting count and code small, per review discussion).

- config: storage_min_free_mb default 0 -> 500.
- Docstring and docs updated to describe the guard as on-by-default and config-tunable.

Testing: unit tests assert the default config now enables the guard and that setting 0 disables it (plus the existing guard behavior tests). Verified end to end on a fresh instance provisioned from this branch: with no config, storage-status reports the 500MB threshold active; a config threshold set above free disk stops all running apps with the storage-too-low message on the next guard check; setting 0 disables enforcement.


zplizzi:
what's the use case for wanting to change this? i feel like setting to enabled and like 500MB is pretty fine as a default? i'm curious why it didn't work in bowei's case.

andrewlaack:
@zplizzi this didn't work in bowei's case because it wasn't enabled by default. This enables it by default and shows this in the ui in case people want to disable it (perhaps during cleanup or if they are severely resource constrained).

zplizzi:
hmm i think i'd be inclined to just enable by default without adding a UI way to enable/disable/set limit. just to keep things simpler both code wise and keeping the amount of configurable settings smaller. if someone really cares, they can always change their instance config file manually and reboot.

PR #238, opened 2026-07-20
Email design doc: multi-tenant-safe send/receive via central broker

andrewlaack:
Adds docs/src/email.md describing how OpenHost instances would send and receive email for their zone, designed to be safe for untrusted multi-tenant users.

Outbound and inbound mail flow through a central OpenHost-operated broker in front of AWS SES. The broker is the trust boundary: instances authenticate with a per-instance Keycloak client credential (mirroring the cert broker), the broker enforces that an instance can only send as its own zone, applies per-instance rate/volume caps, isolates reputation via SES configuration sets, and handles suppression and auto-suspension. Instances never hold AWS credentials.

Sending: instance mailbox server relays as a smarthost client to the broker, which relays to SES SMTP. Receiving: SES to S3 to SNS to a signature-verified webhook fronted by the broker, delivered into the instance's local mailbox. SPF/DKIM/DMARC/MX are auto-published into the instance's own CoreDNS zone at finalize. The mailbox lives on the instance so mail data stays on the operator's zone; a webmail app reads it over a JMAP cross-app service.

Docs only. Added to SUMMARY.md.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -8,6 +8,7 @@
 - [Routing](./routing.md)
 - [User Identity](./user_identity.md)
 - [Persistent Data & Archive](./data.md)
+- [Email](./email.md)
 - [Logs](./logs.md)
 
 # Guides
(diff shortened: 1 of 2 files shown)


andrewlaack:
Superseded by #239, which contains this design doc plus the implemented platform code and a production-readiness checklist, as a single PR.

PR #241, opened 2026-07-21
Add periodic dangling-image pruner thread

andrewlaack:
Every app rebuild re-tags openhost-{app}:latest and orphans the previous image, so untagged (dangling) image layers accumulate over time. Nothing prunes them automatically today: the /api/drop-docker-cache endpoint exists but runs only on explicit operator action, so on long-lived hosts these dangling images pile up and can fill the disk.

This adds a lightweight daemon thread that periodically prunes only dangling images, independent of disk pressure. It runs podman image prune --force (note: no --all), so tagged images for stopped apps are kept and nothing has to be rebuilt on next start. That is what makes an unconditional periodic prune safe.

Details:
- New core/image_pruner.py with the daemon thread, started once per instance from the app bootstrap, mirroring the existing storage guard (idempotent per db_path).
- New containers.prune_dangling_images() helper alongside the existing drop_docker_build_cache(); the difference is the omitted --all flag.
- Config field image_prune_interval_seconds (default 6h; 0 disables). The loop sleeps the interval before its first prune so a mid-build deploy's intermediate layers aren't pruned during the startup rush.
- Prune failures are logged and swallowed so a transient podman error can't kill the loop.

Testing:
- New unit tests for the prune command shape (asserts no --all), the raise-on-error path, the loop swallowing failures, thread start/idempotency/disable, and that the loop sleeps before the first prune. All pass.
- Full compute_space suite green except pre-existing environmental failures (integration and settings_host_prep) that also fail on clean main without the full runtime env.
- ruff check and format clean.
- Vet: opus API plus three agentic runs, all clean.

This is one of two disk-fill mitigation PRs. The other caps journald on-disk size. This PR does not change the storage guard's stop-apps behavior.

### DIFF
--- a/compute_space/src/compute_space/config.py
+++ b/compute_space/src/compute_space/config.py
@@ -75,6 +75,11 @@ class Config:
     # How often (seconds) to prune dangling container images (0 = disabled).
     image_prune_interval_seconds: int
 
+    # Age (seconds) above which a tagged OpenHost app image with no matching app
+    # in the DB is treated as orphaned and pruned (0 = never prune orphaned
+    # tagged images).
+    image_orphan_max_age_seconds: int
+
     ## Ports
     port_range_start: int
     port_range_end: int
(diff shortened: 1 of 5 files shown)


zplizzi (approved):
how will this handle apps that were deleted long ago? would they still have a tag and thus never get cleaned? i think maybe we should additionally do something time based - any images that are say more than a week old (and don't correspond to a current openhost app in any state) can be pruned, even if they still have a tag. otherwise lgtm.

PR #243, opened 2026-07-22
Email: one-NS-record custom domain + mailbox relay-config endpoint

andrewlaack:
Builds on the email foundation branch (andrew/email-dns-records).

Lets an owner use their own mail domain by adding a single NS record delegating it to the instance's CoreDNS.

- email_custom_domain config; the instance serves it as a second authoritative CoreDNS zone (extra Corefile block + zonefile.custom whose NS host is ns.<zone>, so one delegation record suffices) and provisions a second SES identity + SPF/DKIM/DMARC/MX for it.
- custom_domain_delegation_record() + GET /api/email/custom-domain surface the exact NS record to the owner.
- Scoped GET /api/email/relay-config (option b): the mailbox app fetches its SMTP smarthost relay credentials with its app token; the relay password is returned only to the mailbox app (email_mailbox_app_names), never injected into every app's environment.
- docs + tests.

### DIFF
--- a/ansible/templates/config.toml.j2
+++ b/ansible/templates/config.toml.j2
@@ -52,4 +52,13 @@ email_inbound_mx_host = "{{ email_inbound_mx_host | default('inbound-smtp.us-wes
 {% if email_dmarc_rua is defined %}
 email_dmarc_rua = "{{ email_dmarc_rua }}"
 {% endif %}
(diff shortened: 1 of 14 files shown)


andrewlaack:
Superseded by #246, which consolidates all email work into a single PR based on main.

PR #244, opened 2026-07-22
Support git submodules in app deploy and update

zplizzi:
## Why

The upcoming `openhost-minds` app pins its workspace-template code as a (private) git submodule. Today that half-works in a dangerous way:

- **Deploy** clones with `--recurse-submodules`, but the GitHub OAuth token is injected only into the top-level clone URL, so private submodules fail to clone.
- **Update & Reload** (`git_pull`) never touches submodules, so a bumped submodule pointer is fetched into the superproject but never materialized in the working tree — the app "updates" and then rebuilds from the old submodule content with no error.

## What

- `git_pull` now finishes every successful pull/checkout with `git submodule sync --recursive` + `git submodule update --init --recursive` (300s timeout for large submodules; no-op when there is no `.gitmodules`). Newly-added submodules get initialized; bumped pointers get checked out.
- Private-submodule auth uses ephemeral config: `git -c url.https://<token>@github.com/.insteadOf=https://github.com/` on the clone / submodule-update commands. Git exports `-c` config via `GIT_CONFIG_PARAMETERS` to child processes, so recursive submodule fetches inherit it — and the token is never written to any config file.
- Token hygiene: submodule URL sync runs against the restored (clean) origin URL, so relative `.gitmodules` URLs that resolved against a token-bearing origin get scrubbed; on the deploy path the same scrub runs post-clone. `git_pull` logs and returned error strings now redact the token (git error output can echo tokenized URLs).

## Tests

New `test_app_submodules.py` (local file:// fixture repos, same style as `test_set_app_remote.py`): deploy-time clone materializes submodule content; update checks out a bumped submodule pointer; update initializes a submodule added after install; token config formatting. Full compute_space suite (minus podman integration tests): 817 passed, 8 skipped.

Known limitation: relative `.gitmodules` URLs are recommended for private submodules that live in the same org/host as the superproject; absolute private URLs on hosts other than github.com won't authenticate (unchanged from today's top-level behavior).

### DIFF
--- a/compute_space/src/compute_space/core/apps.py
+++ b/compute_space/src/compute_space/core/apps.py
@@ -190,6 +190,18 @@ def inject_github_token_in_url(url: str, token: str) -> str:
     return url
 
 
+def github_token_git_config(token: str | None) -> list[str]:
+    """Ephemeral ``git -c`` args that authenticate GitHub HTTPS fetches.
+
+    Rides in GIT_CONFIG_PARAMETERS, which git propagates to child processes —
+    including recursive submodule clones/fetches — so private submodules
+    authenticate without the token ever being written to a git config file.
+    """
+    if not token:
+        return []
+    return ["-c", f"url.https://{token}@github.com/.insteadOf=https://github.com/"]
+
+
 async def clone_and_read_manifest(
     repo_url: str, github_token: str | None = None
 ) -> tuple[AppManifest | None, str | None, str | None]:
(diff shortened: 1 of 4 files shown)


zplizzi:
Added a second commit: the app test harness's deploy snapshot (_snapshot_working_tree) now includes submodule working trees — git ls-files lists a submodule as a bare gitlink, so a submodule-based app under test previously deployed with the submodule directory empty.

PR #245, opened 2026-07-23
Add /_email/inbound router endpoint to deliver inbound mail to the mailbox app

andrewlaack:
Stacked on andrew/email-custom-domain-and-apps (PR #243).

The email proxy POSTs received mail (raw RFC822, after SNS signature verification and S3 fetch) to https://<zone>/_email/inbound via the imbue-hosted-spaces public door. This router endpoint authenticates that hop in constant time against the per-instance SMTP relay password the instance already holds (Authorization: Bearer = HMAC-SHA256(RELAY_SECRET, zone)), then forwards the request to the first deployed mailbox app (email_mailbox_app_names) on its loopback port at /_email/inbound, which performs actual mailbox delivery. Unconfigured and bad-credential both return 401 (no leak); no mailbox app deployed returns 503. Adds route tests covering auth, no-app, and successful forward.

### DIFF
--- a/ansible/templates/config.toml.j2
+++ b/ansible/templates/config.toml.j2
@@ -58,13 +58,10 @@ email_dmarc_rua = "{{ email_dmarc_rua }}"
 # zone and publishes the same SPF/DKIM/DMARC/MX records into it.
 email_custom_domain = "{{ email_custom_domain }}"
 {% endif %}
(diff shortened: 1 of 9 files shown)


andrewlaack:
Superseded by #246, which consolidates all email work into a single PR based on main.

PR #246, opened 2026-07-23
Email platform support (outbound relay + DNS + custom domain)

andrewlaack:
Platform-side email: opt-in email_* config, CoreDNS SPF/DKIM/DMARC/MX publishing, one-NS-record custom mail domain, and mailbox relay-config (fetched at runtime, no relay secret stored on the instance).

Outbound only: mail relays through the central proxy -> SES. Inbound is NOT handled by this platform code — it is delivered directly to the instance's own mail server (MX -> instance, Stalwart on port 25); the earlier SES -> S3 -> SNS -> proxy -> HTTPS inbound path has been removed.

Design doc in docs/src/email.md. 84 email/relay/dns tests pass; ruff + mypy clean.


andrewlaack:
will add support for the aws flow via our infra and their own smtp server

andrewlaack:
Folded into #248, which is now a single PR against main containing all of this work plus the inbound-always-direct + no-flag changes.

PR #248, opened 2026-07-23
Email support (outbound via SES; inbound always direct to the instance)

andrewlaack:
Complete email support for OpenHost instances, in one PR (supersedes #246, which is folded in here and can be closed).

Outbound: relays through the central proxy to SES, which owns the shared sending reputation and enforces From-domain scope + per-instance rate limits.

Inbound: ALWAYS delivered directly to the instance's own mail server. The zone's MX points at the instance (mail.<zone> A to the instance's public IP) and Stalwart receives on port 25. Inbound never traverses OpenHost infrastructure, so the platform can never read a tenant's mail. There is no SES/managed inbound mode.

No email flag: email is enabled automatically when its prerequisites are present in config (the proxy/frontend URL + per-instance Keycloak client-credentials); when enabled, public_ip is also required for the direct-inbound records. A freshly provisioned instance just has working email once the operator has email infra configured; otherwise it runs without email (no boot failure). Partial config is rejected as a likely typo.

Highlights:
- CoreDNS publishing of SPF/DKIM/DMARC/MX + direct-inbound MX/A, for the built-in zone and delegated custom domains (one NS record).
- SES identity provisioning via the frontend; scoped /api/email/relay-config so only the mailbox app gets the per-instance SMTP credential (fetched at runtime, never baked into config).
- Auto-deploy the mailbox (Stalwart) + webmail (Bulwark) apps when email is enabled.
- Upgrade-safe config loading (tolerates removed email_enabled/email_inbound_* keys; no secret-bearing temp files left behind).
- docs/src/email.md rewritten for the always-direct model.

Tests: ~130 email DNS/config/provisioning/relay/default-apps cases; full non-network suite green (1240 passed); ruff + mypy clean. Vetted heavily across opus, sonnet, and fable (fable caught a severity-5 boot-breaking bug where public_ip was mistakenly treated as an email prerequisite, plus a config upgrade-path break and a temp-file leak — all fixed).


compute_space/src/compute_space/core/email/provision.py
```diff
@@ -0,0 +1,126 @@
...
+    assert config.email_keycloak_issuer_url_resolved is not None
+    assert config.email_keycloak_client_id_resolved is not None
+    assert config.email_keycloak_client_secret_resolved is not None
+    # Inbound is always direct-to-instance, so the MX/A records need the
+    # instance's public IP (validated non-None in Config when email is enabled).
+    assert config.public_ip is not None
+
+    credentials = KeycloakClientCredentials(
+        issuer_url=config.email_keycloak_issuer_url_resolved,
+        client_id=config.email_keycloak_client_id_resolved,
```
zplizzi, line 56: i don't think the keycloak auth stuff should be unique to email? this should be shared for all imbue services.

compute_space/src/compute_space/core/email/provision.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    try:
+        with KeycloakTokenProvider.create(credentials) as token_provider:
+            with EmailProxyClient.create(config.email_proxy_base_url, token_provider) as client:
+                # Built-in zone: the proxy defaults to the caller's zone when no
+                # domain is passed.
+                _provision_zone(
+                    config,
+                    client,
+                    domain=config.zone_domain_no_port,
```
zplizzi, line 67: fyi this stuff will probs need to change a bit to adapt to samuel's multiple-domain work.

compute_space/src/compute_space/core/email/provision.py
```diff
@@ -0,0 +1,126 @@
...
+                    )
+    except EmailProxyError as e:
+        logger.warning(f"Email provisioning skipped: could not reach email proxy: {e}")
+        return
+    except Exception as e:  # noqa: BLE001
+        logger.warning(f"Email provisioning skipped: {e}")
+        return
+
+
+def _provision_zone(
```
zplizzi: nit: i think this could be named more clearly. it's not obvious what "provision zone" means.

compute_space/src/compute_space/core/email/proxy_client.py
```diff
@@ -0,0 +1,103 @@
+"""HTTP client for the email API.
+
+The instance calls the imbue-hosted-spaces frontend (the authenticated public
+door), NOT the private email backend directly — the backend is only reachable
+over Fly's 6PN network. The frontend validates this instance's Keycloak token,
```
zplizzi: nit: details of our backend probs shouldn't be in the open source code

compute_space/src/compute_space/core/dns.py
```diff
@@ -98,6 +98,8 @@ def start_coredns(
     zonefile_path: Path,
     container_gateway_ip: str | None = CONTAINER_GATEWAY_IP,
     coredns_bin: str = "coredns",
+    custom_zone_domain: str | None = None,
```
zplizzi, line 101: i think you'll need to adapt this also to samuel's work on multiple domains

compute_space/src/compute_space/core/dns.py
```diff
@@ -216,12 +331,37 @@ def append_txt_records(zone_file_path: Path, records: list[TxtRecord]) -> None:
...
+    (``_acme-challenge``) or absolute (``_acme-challenge.<zone>.``). Only lines
+    that are TXT records AND owned by that name are considered challenges.
+    """
+    if "IN TXT" not in line:
+        return False
+    owner = line.split(None, 1)[0] if line.split() else ""
+    return owner == _ACME_CHALLENGE_LABEL or owner.startswith(_ACME_CHALLENGE_LABEL + ".")
+
+
 def clear_txt(zone_file_path: Path) -> None:
```
zplizzi: nit: rename to `clear_acme_challenge_records` or something?

compute_space/src/compute_space/web/routes/api/system.py
```diff
@@ -115,6 +163,40 @@ class DropCacheOk:
     output: str
 
 
+@attr.s(auto_attribs=True, frozen=True)
+class EmailRelayConfigResponse:
+    """SMTP smarthost relay config the mailbox app uses to send outbound mail.
+
+    Returned only to the mailbox app (scoped by app name); the relay password is a
```
zplizzi, line 170: i think we should have an openhost "email sending" service, using the same service interface (but provided by openhost, not an app). using SMTP protocol, permissioned same as other services. then any app that wants to send an email can request permission to that service. and the router would then add whatever creds are required and send it along to the imbue service.

compute_space/src/compute_space/config.py
```diff
@@ -57,6 +149,60 @@ class Config:
     #   per-instance client secret (the only sensitive value — treat like the ACME account key)
     cert_api_keycloak_client_secret: str | None
 
+    ## Email
```
zplizzi, line 152: can these comments be made more concise (and 119 chars wide)? and i have some other comments about how we setup the config, will discuss in person.

docs/src/email.md
```diff
@@ -0,0 +1,465 @@
+# Email (Design)
```
zplizzi, line 1: nit: reflow this to 119 chars wide (or for .mds, i'm fine with no hard wrapping also). also, for public docs, can you give this some manual editing to read less like "generated by claude" and be more concise? i would lose some trust in a project if the docs read like they're AI generated.

zplizzi: and remove the details about how our backend works.

zplizzi (changes requested):
let's chat in person about how the auth should get configged - i think this should be shared across all imbue services, and configured dynamically vs in the config file.

compute_space/src/compute_space/web/routes/api/system.py
zplizzi: (which would replace this key-injection-into-specific-apps thing)

andrewlaack:
Superseded by #267 (Email on the shared identity). #267 rebuilds email on the merged shared per-instance identity model, replacing the separate email-only client this PR assumed, and is up to date with main with CI green. Closing this stale branch to avoid confusion.

PR #249, opened 2026-07-23
Add support for local domains

samuelren-imbue:
Not yet supporting .local / mdns usage mode yet. 
Contains docs/local-mode-design.md for WIP notes.


compute_space/src/compute_space/core/domain_store.py
```diff
@@ -0,0 +1,187 @@
...
+    mdns: bool
+    cert_status: str = CERT_STATUS_NONE
+    error_message: str | None = None
+    is_primary: bool = False
+
+    def to_domain(self) -> Domain:
+        return Domain(name=self.name, tls=self.tls, mdns=self.mdns)
+
+
+def _connect(config: Config) -> sqlite3.Connection:
```
zplizzi: we already should have a function for this. also, it's generally better to make funcs take a `db` arg rather than opening a new connection. we don't do this consistently in other core code but should try to move that way. then the caller can reuse an existing connection if it has it (eg all web routes can get a db connection from litestar's dependency injection), and just makes it easier to reason about what things will touch the db.

compute_space/src/compute_space/core/settings_store.py
```diff
@@ -0,0 +1,39 @@
...
+import sqlite3
+from contextlib import closing
+
+from compute_space.config import Config
+
+# Setting keys.
+CLAIM_TOKEN_KEY = "claim_token"
+
+
+def _connect(config: Config) -> sqlite3.Connection:
```
zplizzi: same comment here

compute_space/src/compute_space/config.py
```diff
@@ -33,6 +61,13 @@ class Config:
     # the local port to bind the compute space web server to.
     port: int
 
+    ## Domains
+    # The full set of domains this instance answers on, each with its own scheme
+    # (tls) and discovery method (mdns).  When empty, a single primary Domain is
+    # derived from ``zone_domain`` + ``tls_enabled`` (see ``all_domains``), so
+    # existing single-domain configs are unaffected.  ``domains[0]`` is the primary.
+    domains: tuple[Domain, ...]
```
zplizzi: isn't this in the database now? why is it also in the config? there should only be one source of truth for any piece of info.

docs/local-mode-design.md
```diff
@@ -0,0 +1,667 @@
+# Local (mDNS) mode + concurrent multi-domain serving — design & plan
```
zplizzi: delete this file?

zplizzi (commented):
when yall are happy with this, add me for review again and i'll approve the code-owner stuff

docs/local-mode-design.md
andrewlaack: +1 this file also isn't accurate to the approach used (specifically lines 639-657).

compute_space/src/compute_space/config.py
```diff
@@ -126,11 +161,56 @@ def __attrs_post_init__(self) -> None:
     def zone_domain_no_port(self) -> str:
         return self.zone_domain.split(":")[0]
 
+    @property
+    def all_domains(self) -> tuple[Domain, ...]:
+        """The full domain set — the DB-sourced ``domains`` loaded into the active config by
+        ``rebuild_active_domains``.  Empty only before the first-boot seed runs at startup."""
+        return self.domains
```
andrewlaack: `all_domains` just does `return [self.domains](https://self.domains/)`. If `domains` is empty, it returns empty. It does *not* build anything from `zone_domain`/`tls_enabled` like line 66-67 claims.

andrewlaack: probably just update the comment on lines 66-67

compute_space/src/compute_space/config.py
```diff
@@ -126,11 +161,56 @@ def __attrs_post_init__(self) -> None:
...
+    @property
+    def primary_domain(self) -> Domain:
+        """The canonical domain (``domains[0]``), used by background tasks and outbound links that
+        have no request in hand.  Sourced from the DB, never from the legacy ``zone_domain``."""
+        if not self.domains:
+            raise RuntimeError(
+                "No domains loaded — the DB domain set has not been seeded/rebuilt into the config yet. "
+                "seed_first_boot()+rebuild_active_domains() must run before primary_domain is read."
+            )
+        return self.domains[0]
```
andrewlaack: if `domains` is empty, `primary_domain` raises a RuntimeError, so empty is not "safe with a default" per line 174

compute_space/src/compute_space/web/routes/api/domains.py
```diff
@@ -0,0 +1,232 @@
...
+        # On disk but no longer valid: browsers reject it and renewal is failing — surface it.
+        return CERT_STATUS_ERROR, (record.error_message if record else None) or "certificate expired or unreadable"
+    if record is not None:
+        return record.cert_status, record.error_message  # MISSING: acquiring / error / none
+    return CERT_STATUS_NONE, None
+
+
+def _domain_info(config: Config, domain: Domain, record: DomainRecord | None) -> DomainInfo:
+    name = domain.name_no_port
+    is_primary = name == config.primary_domain.name_no_port
```
andrewlaack: `is_primary` is calculated two ways. The DB column and a string-compare against `config.primary_domain` (lines 97, 214). Since `_domain_list` already loads the records (line 115), just read `is_primary` off the `DomainRecord`. This is  cleaner and avoids the `RuntimeError` `primary_domain` throws on an empty set.

compute_space/src/compute_space/web/routes/api/domains.py
```diff
@@ -0,0 +1,232 @@
...
+    """Restart Caddy off the request path (a daemon thread, after a short flush delay).
+
+    The dashboard is served *through* Caddy (browser → Caddy:443 → router), so restarting Caddy
+    *inside* the request that triggered a domain change would tear down that request's own connection
+    before the response reaches the browser — leaving the UI stale until a manual refresh.  Deferring
+    the restart lets the response (the full domain list) land first.  CoreDNS restarts and cert
+    acquisition stay inline: they're on port 53 / ACME, not the dashboard's HTTP path."""
+
+    def work() -> None:
+        time.sleep(0.5)  # let the response finish flushing through Caddy before we restart it
```
andrewlaack: The deferred Caddy restart leans on a fixed `time.sleep(0.5)` to let the response flush first which is a weak assurance. Better to hook the response-sent event, or at least leave a TODO that it's best-effort.

andrewlaack (commented):
a few notes. Per standup, Zack signed off on the structure so these are just implementation-based comments.

compute_space/src/compute_space/config.py
```diff
@@ -247,8 +333,10 @@ def make_all_dirs(self) -> None:
 
 @attr.s(auto_attribs=True, frozen=True)
 class DefaultConfig(Config):
-    # needs set at runtime, no reasonable default value
-    # zone_domain: str
+    # Legacy field: captured into the DB by the first-boot seed and then never read (see
+    # domain_store/first_boot).  Optional so a post-migration config.toml can omit it — the seed
+    # scrubs the line after capturing it.  The DB `domains` set is the source of truth.
+    zone_domain: str = attr.ib(default="", converter=_lowercase)
```
zplizzi: can we get rid of this? shouldn't the migration run before we try to boot with new code?

compute_space/src/compute_space/config.py
```diff
@@ -126,11 +155,28 @@ def __attrs_post_init__(self) -> None:
     def zone_domain_no_port(self) -> str:
         return self.zone_domain.split(":")[0]
 
+    @property
+    def primary_domain(self) -> Domain:
```
zplizzi: and get rid of this?

compute_space/src/compute_space/web/helpers/zone.py
```diff
@@ -0,0 +1,27 @@
...
+from compute_space.config import get_config
+
+# Scope key under which SubdomainProxyMiddleware stashes the Domain a request
+# arrived on (a ``Domain`` or ``None``).  Lives here — a leaf helper importing
+# only config + litestar — so both the middleware and request handlers can share
+# it without an import cycle.
+ZONE_SCOPE_KEY = "openhost_zone"
+
+
+def zone_for_request(connection: ASGIConnection[Any, Any, Any, Any]) -> Domain:
```
zplizzi, line 13: remove the fallback - the middleware should always run, no? there should only be one way to do things!

compute_space/src/compute_space/core/domain_store.py
```diff
@@ -0,0 +1,190 @@
+"""The DB ``domains`` table is the single source of truth for the hostnames this instance answers on.
```
zplizzi: nits: remove this docstring unless it's actually not obvious info; claude loves to write verbose file level docstrings despite my claude.md instructions not to. rename this file to domains.py, and move the Domain class into here.

compute_space/src/compute_space/core/domain_store.py
```diff
@@ -0,0 +1,190 @@
...
+
+import attr
+
+from compute_space.config import Config
+from compute_space.config import Domain
+from compute_space.config import set_active_config
+from compute_space.core.logging import logger
+
+# Per-domain cert/acquisition status surfaced by /api/domains.
+CERT_STATUS_NONE = "none"  # TLS domain with no cert yet acquired
```
zplizzi: this should be a StrEnum

compute_space/src/compute_space/core/domain_store.py
```diff
@@ -0,0 +1,190 @@
...
+            continue
+        seen.add(host)
+        rows.append((rec.name, int(rec.tls), int(rec.mdns), 0, rec.cert_status, rec.error_message))
+    db.executemany(f"INSERT INTO domains ({_COLS}) VALUES (?, ?, ?, ?, ?, ?)", rows)
+    db.commit()
+    logger.info("Seeded {} domain(s) into the DB", len(rows))
+    return True
+
+
+def seed_domains_from_legacy(config: Config, db: sqlite3.Connection) -> bool:
```
zplizzi: can we move this into the migration? i'd try to avoid having migration/old logic in the core code.

compute_space/src/compute_space/core/domain_store.py
```diff
@@ -0,0 +1,190 @@
...
+    ``<app>.<domain>`` subdomains — or None if none match.
+
+    Longest domain name wins, so overlapping domains resolve to the most specific (e.g.
+    ``host.example.com`` beats a hypothetical ``example.com``).  ``host`` may include a ``:port``; it
+    is compared port-insensitively."""
+    host_no_port = host.split(":")[0].lower()
+    best: Domain | None = None
+    for domain in effective_domains(db):
+        name = domain.name_no_port
+        if not name:
```
zplizzi: nit: put this validation in the Domain class - it's cleaner to validate as early as possible, then consumers don't have to do it individually.

compute_space/src/compute_space/web/middleware/subdomain_proxy.py
```diff
@@ -134,9 +139,19 @@ async def _dispatch(self, scope: Scope, receive: Receive, send: Send) -> None:
...
             return
 
-        app = get_app_from_hostname(netloc)
+        # Resolve which configured Domain this request arrived on, and stash it in
+        # the scope so downstream Litestar handlers (login redirect, cookies,
+        # absolute-URL building) can stay on the arriving domain instead of the
+        # single canonical one.  None for unrelated hosts.
+        config = get_config()
+        with closing(get_db()) as db:
+            zone = match_domain(db, netloc)
```
zplizzi: i think we should 404 (or error in some way) requests that don't match a domain here?

compute_space/src/compute_space/web/middleware/subdomain_proxy.py
```diff
@@ -161,9 +176,15 @@ async def _dispatch(self, scope: Scope, receive: Receive, send: Send) -> None:
         #  - client IP: connection.client is always Caddy; recover the real one
         #    from the X-Forwarded-For Caddy set (see _resolve_forwarded_for).
         # X-Forwarded-Host preserves the original Host so apps that build absolute URLs don't use the proxy's internal hostname.
+        # Proto follows the domain the request arrived on (https for a TLS domain,
+        # http for an mDNS `.local` domain), so an app served on both sees the
+        # right scheme per request rather than one global value.  `app` truthy
+        # implies `zone` is set (get_app_from_hostname only matches a configured
+        # domain); fall back to the primary domain defensively.
+        proto = zone.scheme if zone is not None else config.primary_domain.scheme
```
zplizzi: zone should be non-None by here (see prior comment) so we can drop this fallback.

zplizzi (approved):

zplizzi:
(looks good after feedback is addressed)

compute_space/src/compute_space/core/caddy.py
```diff
@@ -76,13 +142,83 @@ def restart(self) -> None:
...
+                    "reload",
+                    "--config",
+                    str(self.caddyfile_path),
+                    "--adapter",
+                    "caddyfile",
+                    "--address",
+                    self.admin_addr,
+                ],
+                capture_output=True,
+                timeout=30,
```
andrewlaack: reload() runs with timeout=30 but only cold-restarts on returncode != 0. A hung `caddy reload` raises subprocess.TimeoutExpired, which propagates uncaught, contradicting the docstring ("falls back to a cold restart if ... the reload fails") and leaving Caddy on the stale config. Wrap the subprocess.run in try/except subprocess.TimeoutExpired -> _cold_restart_locked(), and add a test for the timeout path.

PR #251, opened 2026-07-24
Link Github on the main page

samuelren-imbue:
Also fixed an issue where the tab placement is slightly different due to having / not having a scroll bar

### DIFF
--- a/compute_space/src/compute_space/core/git_ops.py
+++ b/compute_space/src/compute_space/core/git_ops.py
@@ -1,9 +1,11 @@
+import functools
 import re
 import urllib.parse
 from pathlib import Path
 
 import git
 
+from compute_space import OPENHOST_PROJECT_DIR
 from compute_space.core.util import async_wrap
 
 
@@ -268,7 +270,7 @@ def hard_checkout_ref(repo_path: Path, ref: str) -> None:
     repo.git.clean("-fd")
 
 
-def github_web_url(remote_url: str, branch: str | None) -> str | None:
+def github_web_url_from_remote_url(remote_url: str, branch: str | None) -> str | None:
     """Browsable ``github.com`` URL for ``remote_url`` at ``branch``, or None.
 
     Converts an origin remote (HTTPS, credential-bearing, or SCP-style SSH) into a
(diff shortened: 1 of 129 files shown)


zplizzi:
can you include a screenshot?

samuelren-imbue:
image

compute_space/src/compute_space/core/git_ops.py
```diff
@@ -266,3 +266,55 @@ def hard_checkout_ref(repo_path: Path, ref: str) -> None:
     # checkout -f resets tracked files but leaves untracked files behind, which can
     # shadow modules removed/renamed between revisions. Match origin/ref fully.
     repo.git.clean("-fd")
+
+
+def github_web_url(remote_url: str, branch: str | None) -> str | None:
```
zplizzi: nit: these func names are kinda similar. maybe like `github_web_url_from_git_remote_url` and `github_web_url_from_local_repo` or something?

compute_space/src/compute_space/web/routes/docs.py
```diff
@@ -98,6 +100,11 @@ def _docs_src_dir() -> Path:
 _DEFAULT_INDEX = "introduction"
 _SUMMARY_FILENAME = "SUMMARY.md"
 
+# Resolved once at import: the running checkout's branch/fork is fixed for the
+# process, and the shared nav's "view source" mark (rendered via _nav_header.html)
+# needs it in this route's standalone Jinja environment. None on tarball deploys.
+_SOURCE_URL = get_github_source_url(OPENHOST_PROJECT_DIR)
```
zplizzi: nit: since we're doing this twice, maybe it should live in `git_ops` instead so we can only call it once?

zplizzi (approved):

compute_space/src/compute_space/web/routes/docs.py
samuelren-imbue: caching it in memory has issues since it gets accessed at weird times and gets updated, just keeping as is for now

samuelren-imbue: nvm less of an issue than i thought

PR #254, opened 2026-07-24
add docs for cli http requests

samuelren-imbue:


zplizzi (changes requested):
i think we should see if we can get this auto-generated as OpenAPI spec/docs (but only for the specific routes we want to be part of the "public API", similar to the set you have here). otherwise these docs will get out of date over time / need effort to maintain. better to make things automatic when possible, and litestar has nice support for auto generating OpenAPI specs since we've formally typed all the requests and responses. there's probably some stuff we could do to improve the quality of the generated output by cleaning up docstrings and stuff on those routes though, i'm not sure how that works exactly.

tbh it doesn't need to be perfect tho so a quick pass at generating OpenAPI docs on selected routes and making sure the output looks sane is probs fine.

samuelren-imbue:
I took a look at all the api routes and they seem good to be public? otherwise litestar has a way to disable individual routes in the code

compute_space/src/compute_space/web/routes/pages/apps.py
```diff
@@ -189,4 +189,5 @@ async def add_app(db: sqlite3.Connection, repo: str = "", next: str = "") -> Tem
 pages_apps_routes = Router(
     path="/",
     route_handlers=[dashboard, app_detail, add_app],
+    include_in_schema=False,
```
zplizzi, line 213: are there no routes in here used in the `oh` cli?

zplizzi: i think worth a double check that all routes from the CLI are actually exposed in the yaml

compute_space/src/compute_space/web/app.py
```diff
@@ -227,5 +204,6 @@ def _install_template_globals(app: Litestar) -> None:
             Exception: _log_unhandled_exception,
         },
         on_startup=[_install_template_globals],
+        openapi_config=OPENAPI_CONFIG,
```
zplizzi: this serves a /schema route? is it auth-gated? i'm not sure we want/need this, and if unauthed it leaks info about eg the version being run.

zplizzi (changes requested):
can you see if you can get this to be included in our "manual" (served eg in the docs tab in the instance)? it uses mdbook, claude suggests one way of doing this:

1. Embed Redoc/Swagger UI in a chapter (no plugins, always works)

Put your spec in src/openapi.yaml, then create src/api.md:
```
html
<redoc spec-url="openapi.yaml"></redoc>
<script src="https://cdn.redocly.com/redoc/latest/bundles/redoc.standalone.js"></script>
```

samuelren-imbue (commented):
making some changes

PR #255, opened 2026-07-27
Detect and self-heal rootless-podman overlay build-cache corruption

kilo-gi:
## Problem

On a fresh instance, the initial default-app deploy builds every app **in parallel** (`deploy_default_apps` → `insert_and_deploy` spawns a background build thread per app). Concurrent image builds against one rootless **containers-storage** instance can corrupt the overlay layer metadata — specifically, some layers end up with **empty short-name `link` files**. Any build whose layer chain includes a corrupt layer then fails at mount time:

```
Error: identifier is not a container: preparing container for next step:
mounting new container "...": readlink .../containers/storage/overlay/l: invalid argument
```

(and `readlink .../overlay: invalid argument` at container start). It's **deterministic on retry**, because the corruption is cached. Observed in practice: 2 of 6 default apps (catalog, filestash) stuck in `error` after setup, unfixable by the built-in retry.

Two gaps caused this to be a dead end rather than a self-correcting hiccup:

1. The cache-corruption **detector didn't recognize this pattern**, so the failure was surfaced as a raw error instead of the "drop cache & rebuild" remediation.
2. Nothing **cleared the cache** before retrying — the boot-time serial rebuild and the cross-boot retry both reused the corrupt cached layers.

## Changes

**1. Detection — `core/containers.py`**
`_is_build_cache_corrupt` now also matches `identifier is not a container` and a narrow `readlink .*overlay.*: invalid argument` regex, so the failure is tagged with `BUILD_CACHE_CORRUPT_MARKER`. This both surfaces the dashboard's drop-cache remediation and lands the marker in the app's `error_message` (which #2 keys off).

**2. Self-heal — `core/startup.py`** (no flag)
The boot sweep `check_app_status` now finds apps left in `error` carrying the marker, drops the build cache **once** (global `podman image prune --all --force`), clears their error, and queues them onto the **existing serial** `_restart_apps_sequential` — the one-at-a-time rebuild path that already exists *"to avoid concurrent image builds against the same containers-storage instance."*

Net behaviour: **parallel builds when healthy (fast), prune + serial rebuild on corruption (safe)** — the happy path is untouched.

The corruption signal is read straight from `router.db` (`status='error'` + the marker in `error_message`), which already persists across the restart — no extra state file. Recovery is naturally **one-shot**: it clears each app's error before the serial rebuild, so an app only reappears in the sweep if a *serial*, freshly-pruned rebuild failed with the marker again — which the concurrency that causes the corruption can't produce. A genuinely broken build (bad Dockerfile) fails with a non-marker message and is left alone.

## Testing

- `tests/test_containers.py`: the two real corruption strings (build-time `overlay/l`, start-time bare `overlay`) added to both detection tests, plus innocuous guards (a non-overlay `readlink … invalid argument`; a benign line mentioning `overlay`) so the matcher stays tight.
- `tests/test_check_app_status.py` (5 new): corrupt app → pruned + serially rebuilt + error cleared; non-corrupt error → left alone (no prune); recovered app (marker cleared) → **not** re-pruned on a second sweep; N corrupt apps → **exactly one** prune, all rebuilt; missing repo → skipped.
- Full `compute_space` suite green (699 passed / 32 skipped); ruff + mypy clean.

## Notes / follow-ups

- Recovery fires at the post-setup restart / next boot (when `check_app_status` runs), not mid-setup — in practice seconds later, since setup triggers a restart. A same-session watchdog could make it instant but was left out to keep the change small.
- The deeper prevention (serialize the initial parallel builds, or a global build lock) is intentionally **not** in this PR — the "fast parallel, self-heal on corruption" approach here keeps the happy path fast and was preferred over a flag.
- Verified the root cause end-to-end on an arm64 QEMU VM: repairing the empty `link` files unblocked the builds, and a clean prune + serial rebuild brought catalog and filestash to `running`. (Confirmed the arch defaults in those Dockerfiles are a red herring — podman auto-populates `TARGETARCH` to the native platform; images build `arm64/linux`.)

### DIFF
--- a/compute_space/src/compute_space/core/containers.py
+++ b/compute_space/src/compute_space/core/containers.py
@@ -78,6 +78,10 @@ DEFAULT_CAPABILITIES: frozenset[str] = frozenset(
 _BUILD_CACHE_CORRUPT_FRAGMENTS_UNCONDITIONAL = (
     "storage-driver errored",
     "layer not known",
+    # buildah aborts a build step when the overlay layer it just tried to
+    # mount is unusable (see _OVERLAY_LINK_CORRUPT_RE) and then reports the
+    # discarded build container as "identifier is not a container".
+    "identifier is not a container",
 )
 
 # The "missing local layer blob" error from containers-storage.  The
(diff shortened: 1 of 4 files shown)


kilo-gi:
I ran into this while running in QEMU on ARM, so probably doesn't affect the 90% of users but does make the build succeed for me.

PR #256, opened 2026-07-27
Add option to revoke app permissions

samuelren-imbue:
Lives in detail page


zplizzi (changes requested):
this looks great, can you also show the diff of any other changes in the manifest for the user to approve? stuff like paths becoming public should definitely require approval, and i don't see why we shouldn't just display all changes.

compute_space/src/compute_space/core/apps.py
```diff
@@ -205,6 +206,23 @@ def github_token_git_config(token: str | None) -> list[str]:
     return ["-c", f"url.https://{token}@github.com/.insteadOf=https://github.com/"]
 
 
+async def _remote_ref_is_commit(clone_url: str, ref: str, github_token: str | None) -> bool:
```
zplizzi: nit: move to git_ops

compute_space/src/compute_space/core/apps.py
```diff
@@ -205,6 +206,23 @@ def github_token_git_config(token: str | None) -> list[str]:
...
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
+    if result.returncode != 0:
+        return False
+    return not result.stdout.strip()
+
+
 async def clone_and_read_manifest(
```
zplizzi, line 187: this function is getting pretty complicated and with some duplicated code, can you see if you can restructure maybe with some helper functions to make it logically easier to follow? we could add helpers to git_ops or here depending on which feels more appropriate.

compute_space/src/compute_space/core/apps.py
```diff
@@ -449,6 +497,87 @@ def manifest_newly_declared_permissions_v2(
...
+    group: str
+    label: str
+    old: str
+    new: str
+
+
+# Manifest fields surfaced in the update review, as (attr_name, group, label).
+# Excludes name (app identity, renamed separately), raw_toml, and
+# consumes_services_v2 (diffed by manifest_newly_declared_permissions_v2).
+_REVIEWED_SETTINGS: list[tuple[str, str, str]] = [
```
zplizzi: i don't like this, it'll fall out of date with the manifest type. python has some fun tools for annotating types, i'd try something like this:
```python
class Label:
    text: str

class AppManifest:
    # [app]
    name: Annotated[str, Label("App Name")]
    version: Annotated[str, Label("App Version")]
    ...
```

and then you can pull these out with something like
```python
def manifest_labels():
    hints = get_type_hints(AppManifest, include_extras=True)
    out = {}
    for f in fields(cls):
        hint = hints[f.name]
        for extra in get_args(hint)[1:] if get_origin(hint) is Annotated else ():
            if isinstance(extra, Label):
                out[f.name] = extra.text
    return out
```

you could do like
```
class Label:
    group: str
    label: str
```
if you wanted the categories also.

zplizzi (changes requested):

compute_space/src/compute_space/core/apps.py
zplizzi: also i'd pull this "manifest diff" logic into manifest.py, probably.

samuelren-imbue:
@zplizzi

zplizzi (approved):
sorry for the super slow review on this, all looks great though!

PR #257, opened 2026-07-27
Shared instance identity + Connect to Imbue (instance side)

andrewlaack:
Foundation PR for a single shared per-instance Imbue credential (imbue_identity_*) that authenticates the instance to every Imbue service, plus the instance side of the Connect to Imbue flow.

Managed injection: the ansible template renders the shared imbue_identity_* block, and provisioning injects one credential per instance.

Connect to Imbue (non-managed / recovery): a Settings section starts the flow, the front door redirects back with a one-time code, which is exchanged server-to-server, persisted into config.toml atomically, and the router restarts to load it. Verified live end to end on a real instance.

Awaiting-connect: an instance with the Imbue front door configured but no credential yet boots fine with email off.

Stacked on andrew/email-inbound-direct (#248). ~230 new tests; pre-commit + non-container pytest green.

### DIFF
--- a/ansible/templates/config.toml.j2
+++ b/ansible/templates/config.toml.j2
@@ -22,20 +22,32 @@ start_caddy = {{ 'false' if http_only else 'true' }}
 claim_token_required = true
 data_root_dir = "/home/host/.openhost/local_compute_space"
 apps_dir_override = "/home/host/openhost/apps"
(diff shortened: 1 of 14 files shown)


andrewlaack:
Superseded by #258, which is standalone against main (not stacked on the email branch).

PR #258, opened 2026-07-27
Shared instance identity + Connect to Imbue

andrewlaack:
Introduce a single shared per-instance Imbue credential that authenticates the instance to Imbue services, plus a Connect to Imbue flow. Reworked to use the DB settings table + first_boot.toml introduced in #249 (no config.toml mutation, no restart).

State management: the credential lives in the DB settings table (core/identity_store.py), read live like domains. Managed spaces seed it from first_boot.toml on first boot; non-managed spaces obtain it via Connect. cert-api reads the credential from the store, falling back to the deprecated cert_api_keycloak_* config fields for already-deployed instances.

Connect to Imbue: a Settings section sends the owner to Imbue to authorize; Imbue returns a one-time code, which the instance exchanges server-to-server and writes to the settings table. No config file mutation, no restart (read live). Verified live end to end on a real instance: the credential lands in the settings table and the status endpoint flips to connected with no restart.

Around 113 new tests plus 50+ live edge cases on a real instance; pre-commit and test CI green.

Depends on #249 (merged).

### DIFF
--- a/compute_space/src/compute_space/core/connect.py
+++ b/compute_space/src/compute_space/core/connect.py
@@ -1,0 +1,83 @@
+"""Instance side of the "Connect to Imbue" flow.
+
+Managed spaces get their shared Imbue credential injected at provision time. A
+non-managed (self-hosted) instance instead obtains it here: the owner clicks
+"Connect to Imbue" in Settings and is sent to Imbue to authorize; Imbue returns a
+one-time code to this instance. This module builds the authorization URL, exchanges
+the code for the credential, and persists the credential into the instance's
+config.toml so a restart picks it up through the normal config path
+(Config.instance_identity).
+"""
+
+from __future__ import annotations
+
+from urllib.parse import urlencode
+
+import attr
+import httpx
+
+from compute_space.core.tls.keycloak import KeycloakClientCredentials
+
+# The instance-side callback path Imbue returns the one-time code to (?code=).
+CONNECT_CALLBACK_PATH = "/api/settings/connect-imbue/callback"
+
+_CONNECT_START_PATH = "/connect/imbue"
+_CONNECT_EXCHANGE_PATH = "/connect/imbue/exchange"
+
+
+@attr.s(auto_attribs=True, frozen=True)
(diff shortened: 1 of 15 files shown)


compute_space/src/compute_space/core/connect.py
```diff
@@ -0,0 +1,115 @@
...
+    ``instance_base_url`` is this instance's own https origin; the one-time code is
+    returned to it (at CONNECT_CALLBACK_PATH), so the callback must be on the
+    instance's own zone.
+    """
+    callback = f"{instance_base_url.rstrip('/')}{CONNECT_CALLBACK_PATH}"
+    query = urlencode({"zone": zone, "callback": callback})
+    return f"{frontend_base_url.rstrip('/')}{_CONNECT_START_PATH}?{query}"
+
+
+def exchange_code_for_credential(
```
zplizzi, line 50: what's the one time code from?

compute_space/src/compute_space/core/connect.py
```diff
@@ -0,0 +1,115 @@
...
+        return KeycloakClientCredentials(
+            issuer_url=str(body["issuer_url"]),
+            client_id=str(body["client_id"]),
+            client_secret=str(body["client_secret"]),
+        )
+    except (ValueError, KeyError, TypeError) as e:
+        raise ConnectError("connect exchange returned a malformed response") from e
+
+
+def persist_instance_identity(config_path: str, credential: KeycloakClientCredentials) -> None:
```
zplizzi: i've been trying to avoid making the config file mutable, i feel like that gets messy. can we store these in the db instead? the initial secret from our provisioning can go into the `first_boot.toml` samuel added here: https://github.com/imbue-openhost/openhost/pull/249, and he also added a `settings` table for mutable key/value pairs.

compute_space/src/compute_space/core/connect.py
andrewlaack: It's the single-use code we issue to an instance's callback after the owner authorizes in the frontend

PR #259, opened 2026-07-27
add bot tag to not index pages

samuelren-imbue:
Use this tag: <meta name="robots" content="noindex"> to remove page from major search providers.


samuelren-imbue:
deindexes main pages but not apps, which are hopefully inaccessible? and should be up to the apps

PR #264, opened 2026-07-30
Forward original Host header to proxied apps

kilo-gi:
## What

The subdomain proxy stripped the inbound `Host` header and relied solely on `X-Forwarded-Host`, so backend apps saw the `Host: 127.0.0.1:<port>` that httpx synthesizes from the internal target URL.

Many frameworks only read `Host` and ignore `X-Forwarded-Host` unless explicitly configured — Django `ALLOWED_HOSTS`, absolute-URL builders, cookie-domain logic, CSRF/origin checks. This has bitten us at least twice with apps that don't respect `X-Forwarded-Host` out of the box.

## Change

Preserve the original public `Host` on the HTTP forward path (while still sending `X-Forwarded-Host`, so apps that read either keep working). This matches the default behavior of every mainstream reverse proxy (nginx `proxy_set_header Host $host`, Caddy `reverse_proxy`, Traefik) — it's not a protocol violation, it's the norm.

Notes:
- **httpx** respects an explicit `Host` header and does not duplicate it (verified).
- **WebSocket path is intentionally left unchanged**: the `websockets` client (v16) sets its own `Host` at handshake time and its `Headers.update()` *appends* rather than replaces, so injecting a `Host` there would produce a duplicate and could break the handshake. WS backends validate `Origin`, not `Host`, so status quo is correct.

## Tests

- Added an assertion to `test_proxy_forwards_headers` that the backend sees `Host: test-app.<zone>:<port>` rather than the internal loopback address.
- Ruff, mypy, and the `test_subdomain_proxy` unit tests pass locally. The integration test is container-gated, so it runs in CI/Linux.

### DIFF
--- a/compute_space/src/compute_space/tests/test_integration.py
+++ b/compute_space/src/compute_space/tests/test_integration.py
@@ -1096,6 +1096,9 @@ class TestContainerE2E:
         assert "x-forwarded-for" in headers_ci
         assert "x-forwarded-host" in headers_ci
         assert headers_ci.get("x-forwarded-proto") == "http"  # tls_enabled is False in tests
+        # Host is preserved as the public hostname, not the internal 127.0.0.1:<port>,
+        # so apps that only read Host (Django ALLOWED_HOSTS, absolute-URL builders) work.
+        assert headers_ci.get("host") == f"test-app.{config.zone_domain}:{config.port}"
 
     def test_proxy_forwarded_headers_trust_model(self, admin_session, config):
         """X-Forwarded-Proto/Host are derived by the router, never taken from the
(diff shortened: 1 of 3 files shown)


zplizzi (approved):
moved the location to where we do other header stuff, so it gets included in websocket too. this looks good tho, thanks for fixing - idk why i didn't get this right the first time.

kilo-gi:
ok, Here's the code that shows setting 'Host' in additional headers on a websocket ends up with it duplicated:

Library sets its own Host from the dial target `ClientProtocol.connect()`: https://github.com/python-websockets/websockets/blob/16.0/src/websockets/client.py#L109
Our additional_headers merged via `.update()`: https://github.com/python-websockets/websockets/blob/16.0/src/websockets/asyncio/client.py#L100
`Headers.__setitem__()` (what update uses) appends instead of replacing: https://github.com/python-websockets/websockets/blob/16.0/src/websockets/datastructures.py#L109-L111

I'm going to put in a comment, then look a little further into why this is so hard on the websocket side of things.

PR #267, opened 2026-07-31
Email on the shared identity: outbound via a router-mediated 'email' service (SMTP), inbound direct

andrewlaack:
Adds email to OpenHost instances on the shared Imbue identity. Inbound is delivered directly to the instance's mail server (MX -> mail.<zone> -> instance IP, Stalwart on :25), never touching Imbue infra. Outbound is a router-mediated service.

Outbound as an OpenHost service:
- Email is an "email" v2 service (github.com/imbue-openhost/openhost/services/email), permissioned like any other service. Apps request the "send" grant in their manifest.
- The router runs a local SMTP submission listener (loopback + container gateway only, never public). Apps relay to it with SMTP AUTH (username = app name, password = their OPENHOST_APP_TOKEN).
- The router checks the grant, attaches the per-instance relay credential, and forwards to the Imbue SES smarthost. The app never sees the relay credential.
- Replaces the old app-scoped path: the /api/email/relay-config route and email_mailbox_app_names allowlist are removed.

Other pieces:
- CoreDNS publishing of SPF/DKIM/DMARC + direct-inbound MX/A; scoped ACME-challenge clear so cert renewal never wipes email records.
- Startup provisioning of the SES identity + DKIM (best-effort, no-op when email off).
- Mailbox (Stalwart) + webmail (Bulwark) auto-deployed when email is enabled. Owner-facing /api/email/custom-domain (BYO domain via one NS record).
- Carries system-agent migration v9 (moves any pre-existing cert_api_keycloak_* config into the settings table); belongs to the shared-identity work, included here because email depends on it.

Companion (must land in lockstep): openhost-stalwart-email-server#12.

Verification: unit tests (auth, grant, SMTP relay path) + full suite + ruff/mypy + vet clean; 50 live full-process e2e; real Stalwart-path with confirmed inbox delivery.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ openhost up --dev
 # (for example: http://my-app.lvh.me:8080)
 openhost up --dev --zone-domain lvh.me
 
+
 # check prerequisites
 openhost doctor
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,6 @@ openhost up --dev
 # (for example: http://my-app.lvh.me:8080)
 openhost up --dev --zone-domain lvh.me
 
-
 # check prerequisites
 openhost doctor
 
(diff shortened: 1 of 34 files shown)


andrewlaack:
Note: the identity-migration is related to the shared-identity work, not email. It is here though because this pr needs that migration. This can be split into another pr if needed.

PR #268, opened 2026-08-03
Fix test suite warnings

samuelren-imbue:
Address the 2000+ warnings during tests. 
bump up litestar requirement
jinja engine import path
explicit imports for litestar
to_asgi_response turn "app" kwarg to positional
switch from passing cookies as kwarg to setting in client
mute PRAGMA-foreign-keys lint

### DIFF
--- a/compute_space/src/compute_space/tests/test_versioned_migrations.py
+++ b/compute_space/src/compute_space/tests/test_versioned_migrations.py
@@ -503,8 +503,6 @@ def _scan_migration_for_unsafe_ops(migration: Migration) -> list[str]:
     (the PRAGMA is a no-op inside a tx, which tends to silently defeat
     the author's intent). Extend as we encounter more gotchas.
     """
-    if migration.allow_pragma_foreign_keys:
-        return []
     findings: list[str] = []
     if isinstance(migration, SqlFileMigration):
         sql_path = Path(inspect.getfile(migration.__class__)).resolve().parent / migration.sql_file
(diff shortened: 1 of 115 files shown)


compute_space/src/compute_space/tests/test_versioned_migrations.py
```diff
@@ -519,6 +517,8 @@ def _scan_migration_for_unsafe_ops(migration: Migration) -> list[str]:
         findings.append(
             f"{type(migration).__module__}.{type(migration).__name__}: PRAGMA foreign_keys inside Python migration"
         )
+    if migration.allow_pragma_foreign_keys:
```
kilo-gi: nit: why not check this at 509/516 instead of removing after the fact?

kilo-gi (approved):
I assume NamedDependency and FromPath / FromQuery are like annotations on the types and that they all function the same and will convert back into the underlying type transparently? And Litestar wants them so it can doublecheck it's populating them all correctly? given yes and yes, all that LGTM.

Test code all looks simple - just setting the cookies a slightly different way.

I do think I find it easier to think about the findings list as 'we only ever add to this list'. that would prevent a weird bug where we were testing a line that had the 'PRAGMA foreign_xyz' magic string as content instead of as an SQL statement, but that would be such a strange test so I don't think this is actually incorrect. So it's a nit.


A little bit more background in the description (here's the edits I made, here's the error it fixes, here's why) would make this easier/nicer to review (since I don't have litestar details memorized), but you don't have to fix that before you merge, just would be nice for next time.

zplizzi (approved):
lgtm!

PR #270, opened 2026-08-06
Keep the dashboard responsive during self-update

andrewlaack:
Keeps the dashboard responsive during a self-update instead of showing a dead socket, and takes compute_space fully down for the apply so migrations run against a quiet system.

Previously the update ran with the old compute_space still serving: it checked out new code, ran migrations and installed dependencies underneath the running process, then restarted at the end. That meant migrations had to reason about a live reader of the router DB and data dir, and `pixi install` mutated the environment of the running interpreter.

Now the apply detaches into its own transient systemd unit, stops openhost for the whole walk, and starts it again at the end. A detached updater covers the window on 80/443 with the on-disk cert, serving the live progress log the whole time.

Because the apply is a child of openhost's cgroup, it cannot be the thing that stops openhost — the stop would kill it mid-walk. So `update apply` always re-launches itself as `openhost-apply.service` before doing anything. That unit resets the start-rate-limit and starts openhost from `ExecStopPost`, so every exit path — success, failure, crash, OOM, SIGKILL, `RuntimeMaxSec` — still brings the instance back up. There is deliberately no inline fallback when systemd-run is missing: refusing to update is better than stopping a service nothing can start.

Fetching and the preflight checks happen before the stop, since they only read the repo. A dirty tree or an unreachable remote therefore costs no downtime at all, and the fetch — the step most likely to hang — runs with the instance still serving.

`update apply` returns as soon as systemd accepts the job, so the progress log is the result channel; `--wait` blocks and turns the walk's outcome back into an exit code for scripts.

Caddy is the only TLS terminator and, like CoreDNS, is an in-process child of compute_space, so both go down with it. The updater runs as its own transient unit, binds 80/443 while openhost is stopped, and hands the ports back before the new Caddy starts.

CoreDNS is not covered, and that is the significant remaining cost. The instance is the only authoritative nameserver for its own zone, so while it is down a visitor whose resolver cache expires gets a DNS failure rather than the update page. Measured on a live instance during a ~155s walk with the old 60s TTL: queries straight at the instance failed for the whole window, and 8.8.8.8 returned SERVFAIL or timed out for 65% of samples.

The zone TTL is raised from 60s to 300s here to shrink that. A resolver re-queries once per TTL, so the chance it needs to during an outage of length W is roughly W/TTL: for a real update of a few seconds that drops from about 1-in-7 resolvers to 1-in-33, and it helps every router restart rather than only updates. It is a mitigation, not a fix — a resolver with no cached entry still fails, and a long multi-hop walk still exposes about half of them. ACME is insulated from the change: challenge TXT records get an explicit 60s TTL, so a renewal cannot be served the previous run's token, and the SOA minimum stays 60 so negative caching is unchanged. Verified against real CoreDNS: the wildcard answers with TTL 300 and `_acme-challenge` with 60.

Removing the outage entirely, rather than making fewer people hit it, means extracting CoreDNS and Caddy into their own units — the next change.

Live testing on Hetzner instances drove 90+ real updates and found five defects that unit tests could not (the transient unit had no `$HOME`, so no update completed at all; the start-rate-limiter refused the failsafe; `start` was a silent no-op on an already-active unit; an interrupted walk left the page polling forever; and an orphaned updater could hijack a later unrelated restart). All are fixed here.

The update to this patch is still the old flow; every update after it uses the new one.

No migration is needed to adopt it. Nothing here writes host state — no unit files, no /etc, no sysctls — and the zone file is regenerated at every boot, so the new TTL applies as soon as the instance restarts. Verified end to end from a real release: a v0.1.2 instance (host migration version 1, `$TTL 60`, no `--wait`) was updated to this branch from the dashboard. The upgrade hop ran on the old flow as expected, with the app served throughout and no detached unit; afterwards the host was on this code with the ledger advanced 1 -> 9, host version matching, `$TTL 300`, and `--wait` available. The next update on that same instance then used the new flow: openhost inactive for the whole ~149s walk, `openhost-apply.service` and `openhost-updater.service` both active, 67 samples of the update page, and a clean return to the dashboard.

The one thing that makes the upgrade hop work is `start_openhost` using `restart` rather than `start`: the old flow never stops openhost, so a `start` would be a silent no-op on an active unit and the freshly installed code would never boot.

```mermaid
sequenceDiagram
    participant B as Browser
    participant CS as compute_space plus Caddy and CoreDNS
    participant AP as openhost-apply unit
    participant U as openhost-updater unit
    participant SD as systemd

    B->>CS: POST /api/settings/update
    CS-->>B: 200 with token, open /updating
    CS->>SD: launch openhost-apply, detached, after response
    SD->>AP: start

    AP->>U: launch, wait for ready
    AP->>SD: systemctl stop openhost
    SD-->>CS: SIGTERM cgroup, Caddy and CoreDNS exit

    rect rgb(255,235,235)
        Note over B,U: openhost down for the whole apply, updater owns 80/443
        U->>U: bind 443/80, serve page and full logs
        loop apply, with nothing of ours holding the DB or data dir
            AP->>AP: fetch, checkout, migrate, pixi install
            B->>U: poll /updates, token unlocks logs
        end
        AP->>SD: systemctl start openhost
        SD->>CS: start new compute_space
        CS->>U: stop updater, release ports
        CS->>CS: new Caddy and CoreDNS start
    end

    CS-->>B: dashboard back, /updating redirects to /settings
    Note over AP,SD: ExecStopPost always starts openhost, so a crashed apply still recovers
```

## What it looks like

Both captured from a live instance mid-update, while openhost is stopped and the detached updater owns 80/443.

The update page as the owner and as an anonymous visitor

The owner's tab holds the one-time update token, so it streams the progress log. Every other hostname on the instance gets the same page from the updater but no log, because `/updates` answers 403 without that token. Non-browser requests get `503` with `Retry-After` instead of the page, so an app's XHR or a webhook never tries to parse HTML as JSON.

Layout:
- openhost_system_agent/updater: the detached server, its launcher, the progress log, and the shared on-disk paths (token, progress, certs)
- openhost_system_agent/update.py: self-detaches into openhost-apply.service, then stops openhost before touching the working tree
- compute_space: token-returning /api/settings/update, owner-authed /updates for live logs, the /updating page, and the port handoff back to Caddy


andrewlaack:
will update this to stop / start openhost instance at start and end of updating for simplicity and to support detacthing the caddy / coredns servers from the openhost systemd service.

compute_space/src/compute_space/web/routes/api/settings.py
```diff
@@ -149,10 +195,38 @@ async def apply_update() -> None:
...
+
+@attr.s(auto_attribs=True, frozen=True)
+class UpdateProgressResponse:
+    entries: list[dict[str, Any]]
+    terminal: bool
+
+
+@get("/updates", guards=[require_owner_auth])
+async def update_progress() -> UpdateProgressResponse:
+    # Same path+shape the detached updater serves during the final restart; this
```
zplizzi: there's no "long apply phase" here anymore, right? i wonder if we even need this update-tracking logic here. i guess no harm.

compute_space/src/compute_space/core/system_agent/update_token.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+import secrets
+
+from compute_space.core.logging import logger
```
zplizzi, line 5: nit - this file and update_progress.py feel like they're system agent related? seems like this could go in system_agent.py or maybe all 3 files should go in a system_agent subfolder. feel free to ignore tho

compute_space/src/compute_space/web/start.py
```diff
@@ -64,6 +67,9 @@ def _terminate_children(children: list[subprocess.Popen[bytes]]) -> None:
 def _bootstrap(config: Config) -> None:
     """One-time process-wide initialization shared by the setup and full apps."""
     set_active_config(config)
+    # Point the shared updater paths at this instance's data dir so compute_space,
+    # the agent, and the detached updater all use the same progress log / token.
+    os.environ[DATA_DIR_ENV] = str(config.openhost_data_path)
```
zplizzi, line 76: this feels weird, we're just trying to get this into calls to the system updater? can we just inject it when we call that command?

compute_space/src/compute_space/web/templating.py
```diff
@@ -0,0 +1,58 @@
+"""Jinja environments that are pinned to the code the process booted with.
```
zplizzi: can we drop this now?

openhost_system_agent/src/openhost_system_agent/updater/progress.py
```diff
@@ -0,0 +1,136 @@
...
+            pass
+
+
+def _ensure_updater_dir() -> None:
+    d = updater_dir()
+    d.mkdir(parents=True, exist_ok=True)
+    _chown_to_host(d)
+
+
+PHASE_DONE = "done"
```
zplizzi: nit: make a StrEnum

compute_space/src/compute_space/web/routes/api/settings.py
zplizzi: or does this display the full logs after the update completes also? that might be nice i guess

zplizzi (approved):
lgtm! thanks for all the work here, this is really tricky stuff. i do wonder if it would be simpler to just install the updater as a permanent systemd service in ansible in this PR instead of in a future PR - i think this could simplify some of the logic around re-parenting and installing a temporary systemd service. but i'm fine with this also. would need to figure out a different channel for the compute space and the updater service to communicate.

zplizzi:
also, can you make a doc somewhere (in repo or in our notes vault) with any useful notes about how the updater works, and including your diagram? it's really helpful, i'd like to keep it. you can just include a link to a diagram in `https://mermaid.live` for now.

and also with a sketch of the changes we'd like to make to this in the future to improve it further.

andrewlaack:
added a doc in the vault at andrew_docs/updating_ux

PR #271, opened 2026-08-07
Add mdns support

samuelren-imbue:
Custom mdns listener serving a wildcard domain. Works natively on mac and with configs on Linux. 

Windows can set *.openhost.local dns -> specific address, so CoreDNS serves the domain as well, but untested. 

Setting public domain to local instance also untested.


samuelren-imbue (commented):
making some changes, mostly reducing comments and doing some consolidation. 
one other change is validating ips before answering mdns queries.

samuelren-imbue:
@zplizzi

openhost_system_agent/src/openhost_system_agent/migrations/versions/v0002_baseline.py
```diff
@@ -79,7 +79,10 @@ def build_openhost_service_unit(host_uid: int) -> str:
         f"Environment=XDG_RUNTIME_DIR=/run/user/{host_uid}\n"
         f"Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/{host_uid}/bus\n"
         + RECLAIM_EXEC_START_PRE
-        + "ExecStart=/home/host/.pixi/bin/pixi run python -m compute_space\n"
+        # --as-is (= --no-install --frozen) makes startup fully offline: no lock-vs-manifest check,
+        # no install/build, so a reboot during a WAN/DNS outage can't fail fetching build deps or
+        # abort on a merely-stale lock file (which --locked would).
+        + "ExecStart=/home/host/.pixi/bin/pixi run --as-is python -m compute_space\n"
```
zplizzi: is this necessary?

openhost_system_agent/src/openhost_system_agent/migrations/versions/v0007_seed_domains_and_scrub.py
```diff
@@ -69,7 +69,7 @@ def _seed_domains(db: sqlite3.Connection, openhost: dict[str, object]) -> None:
         return  # nothing to capture (e.g. already-scrubbed config)
     tls_enabled = bool(openhost.get("tls_enabled", False))
     seen = {zone_domain.split(":")[0]}
-    rows: list[tuple[str, int, int, int, str, None]] = [(zone_domain, int(tls_enabled), 0, 1, "none", None)]
+    rows: list[tuple[str, int, int, str, None]] = [(zone_domain, int(tls_enabled), 1, "none", None)]
```
zplizzi: this file needs changed?

.pre-commit-config.yaml
```diff
@@ -22,6 +22,12 @@ repos:
         language: system
         pass_filenames: false
         types: [python]
+      - id: pixi-lock
```
zplizzi, line 27: try `requires-pixi = "whatever version you want"` in the pyproject.toml and then just doing `pixi lock`?

ansible/setup.yml
```diff
@@ -7,6 +7,10 @@
 # Optional variables:
 #   -e public_ip=1.2.3.4         (defaults to target IP)
 #   -e initial_user=ubuntu        (defaults to root)
+#                                 A `.local` domain (e.g. -e domain=openhost.local)
```
zplizzi: this seems like an out of place comment? doesn't relate to the line before or after

compute_space/src/compute_space/core/mdns.py
```diff
@@ -0,0 +1,631 @@
...
+def _start_and_register(bases: tuple[str, ...], lan_ip: str, lan_ip6: str | None) -> None:
+    try:
+        set_active_mdns(start_mdns(bases, lan_ip, lan_ip6))
+    except OSError as exc:
+        # A :5353 bind clash (e.g. an avahi/systemd-resolved without SO_REUSEPORT) must never take
+        # the router down — degrade to no mDNS and keep serving.
+        logger.warning("mDNS responder failed to start ({}); continuing without it", exc)
+
+
+def ensure_mdns_for_domains(db: sqlite3.Connection, lan_ip: str | None = None, lan_ip6: str | None = None) -> None:
```
zplizzi: nit: i think you can make this file not require a db handle - you can just pass in `effective_domains` directly? that makes it clearer to the caller how it works, and easier to test.

compute_space/src/compute_space/web/start.py
```diff
@@ -140,23 +145,32 @@ def main() -> None:
...
                 coredns_bin=_ensure_coredns_binary(config),
+                lan_ip=lan_ip,
+                lan_ip6=lan_ip6,
             )
             # Register so /api/domains can regenerate zones + restart CoreDNS when a domain is added.
             set_active_coredns(coredns)
 
+        # Start the wildcard mDNS responder if any `.local` domain is configured (zero-config LAN
+        # discovery alongside CoreDNS); reconciled here and by /api/domains, so it toggles at runtime.
+        ensure_mdns_for_domains(db, lan_ip=lan_ip, lan_ip6=lan_ip6)
```
zplizzi: so here you'd use the db to get `effective_domains` instead of passing in a handle to the db. and i'd rather if there's a clear `if` guard on this - if there's no .local domains, we should never call into the mdns code. that makes it so a reader of this file doesn't need to go into mdns code to figure out if there's any unexpected side effects etc of this call.

compute_space/src/compute_space/core/util.py
```diff
@@ -2,14 +2,90 @@
...
 import os
+import socket
 from collections.abc import Callable
 from collections.abc import Coroutine
 from pathlib import Path
 from typing import Any
 
 import attr
 
+_PROBE_TARGET = {
```
zplizzi: the new logic in this file seems complicated. i don't fully understand it but can we simplify this? can we just do
```
ip route get 1.1.1.1     # shows src <your IPv4> dev <iface>
ip route get 2606:4700:4700::1111
```
to get the ip4 and ip6 LAN addr?

compute_space/src/compute_space/web/start.py
```diff
@@ -183,6 +197,17 @@ def main() -> None:
                 "A TLS domain is configured but start_caddy is False. Caddy is required for TLS termination."
             )
 
+        # The edge is up now, so the IPv6 reachability probe can finally succeed.  Re-read before
+        # arming the watcher, or its first poll reads the pre-Caddy snapshot as a move and restarts.
+        published = lan_addresses()
```
zplizzi: this block of code also feels more complicated than necessary. if we can get out LAN addr before caddy is up, can we fold all this into the if clause where `ensure_mdns_for_domains` gets called? and the IP watcher would only get started if you have a .local addr which seems like what we'd want.

compute_space/src/compute_space/core/dns.py
```diff
@@ -93,12 +99,23 @@ def _coredns_bind_ip(public_ip: str) -> str:
...
-        return public_ip
+    return default_route_source_ip() or public_ip
+
+
+# The http edge (Caddy) serves `.local` on plain http; an address it doesn't answer on is one we
+# must not publish.
+_EDGE_HTTP_PORT = 80
+
+
+def publishable_lan_ip6() -> str | None:
```
zplizzi: this should move alongside the other IP related logic i think?

compute_space/src/compute_space/core/dns.py
```diff
@@ -129,22 +146,45 @@ def public_dns_zones(config: Config, db: sqlite3.Connection) -> tuple[DnsZone, .
             zonefile_path=config.coredns_zonefile_path_for(d.name_no_port, d.name_no_port == primary_no_port),
         )
         for d in effective_domains(db)
-        if not d.mdns
     )
 
 
+def _publishable_zones(zones: tuple[DnsZone, ...], lan_ip: str | None) -> tuple[DnsZone, ...]:
```
zplizzi: when would you not have a LAN IP?

zplizzi (changes requested):
looks good overall but i'd like you to try to tighten up / simplify the logic around the IP addresses and the start.py. i think this is making non-mdns code more confusing to understand which i don't like.

.pre-commit-config.yaml
samuelren-imbue: this works for verifying pixi in the actual deploy, but ci still requires a pixi from a fresh environment and doesn't reference pyproject. moved regen to script.

compute_space/src/compute_space/core/util.py
samuelren-imbue: this is essentially the same mechanism, but just done in python. most of it is just to filter the returned output and do checks on them. 

the logic is a bit weird and there was a fallback that was put in. im gonna separate it into public_ip and lan_ip and make the caller sites figure out which to use to be more explicit.

compute_space/src/compute_space/core/dns.py
```diff
@@ -89,21 +81,43 @@ def _host_upstream_resolvers() -> list[str]:
     return resolvers or list(_FALLBACK_UPSTREAM_DNS)
 
 
-def _coredns_bind_ip(public_ip: str) -> str:
-    """Return the local address CoreDNS should bind for authoritative DNS.
+def _coredns_bind_ips(private_ip: str | None, private_ip6: str | None, public_ip: str) -> tuple[str, ...]:
```
zplizzi: nit: this name sounds like a verb/action

compute_space/src/compute_space/core/dns.py
```diff
@@ -89,21 +81,43 @@ def _host_upstream_resolvers() -> list[str]:
...
+        )
+    return bindable
+
+
+# The http edge (Caddy) serves `.local` on plain http; an address it doesn't answer on is one we
+# must not publish.
+_EDGE_HTTP_PORT = 80
+
+
+def publishable_private_ip6() -> str | None:
```
zplizzi: TODO

PR #272, opened 2026-08-07
docs: copy any page, or the entire manual, as markdown

CarlKho-Minerva:
Follow-up to the community-chat ask — a way to hand the manual to an AI agent in one shot.

- `/docs/<page>.md` — one page's markdown source
- `/docs/all.md` — every page, concatenated in `SUMMARY.md` order
- copy icon beside each page heading, and beside the sidebar title for the whole manual

image

Markdown is passed through verbatim, so the export can't drift from what `/docs/<page>` serves. Each icon is a plain link to the markdown it copies, so it degrades rather than breaking without JS or a secure context. 5 tests added.

### DIFF
--- a/compute_space/src/compute_space/web/routes/docs.py
+++ b/compute_space/src/compute_space/web/routes/docs.py
@@ -103,6 +103,8 @@ def _docs_src_dir() -> Path:
 _DEFAULT_INDEX = "introduction"
 _SUMMARY_FILENAME = "SUMMARY.md"
 
+_ALL_MARKDOWN_PATH = "/docs/all.md"
+
 
 # ─── Space navigation header ────────────────────────────────────────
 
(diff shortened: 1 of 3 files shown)


zplizzi:
looks good to me at a quick glance, @samuelren-imbue can you give a full review and especially make sure there's no way for this to load arbitrary files off the instance that aren't docs.

zplizzi:
fwiw, this would be simpler if the docs rendering were client side instead of server side.. to consider for the future

compute_space/src/compute_space/web/routes/docs.py
```diff
@@ -103,6 +103,9 @@ def _docs_src_dir() -> Path:
 _DEFAULT_INDEX = "introduction"
 _SUMMARY_FILENAME = "SUMMARY.md"
 
+_MARKDOWN_SUFFIX = ".md"
```
samuelren-imbue: nit: this is used in one function but ".md" is inlined elsewhere. either inline that or make all places use this

docs/src/introduction.md
samuelren-imbue, line 1: @zplizzi up to you, but do we want to explicitly talk about AI dev support in the manual? my impression was that it should not be a primary feature but we already have claude.md.

samuelren-imbue (commented):
LGTM other than the minor stuff

docs/src/introduction.md
zplizzi: seems ok to me

zplizzi:
(btw CI will fail for external PRs - we will fix that separately)

compute_space/src/compute_space/web/routes/docs.py
CarlKho-Minerva: inlined it. ".md" was already inline in _resolve_doc_path and _slug_from_href, so this matches the rest of the file. pushed in 96ddfb7

docs/src/introduction.md
CarlKho-Minerva: happy to trim it down or drop it if you'd rather keep the manual focused.

CarlKho-Minerva:
on loading arbitrary files: both /docs/<page> and /docs/<page>.md go through the existing _resolve_doc_path, and the .md suffix is stripped before the slug regex runs, so ../secrets.md and foo.md.bak stay 404s

symlinks were the case i wasn't sure about, so i checked: dropped evil.md -> /etc/passwd and evil.md -> ../secret.txt inside docs/src/, both 404, since resolve() + relative_to(src) catches them. /docs/all.md only walks SUMMARY.md entries and resolves each one the same way, so a symlinked page listed in SUMMARY gets skipped too

no new filesystem access path, it all funnels through the same gate /docs/<page> already used

and noted on the CI, thanks @zplizzi @samuelren-imbue 🙌🎉

zplizzi (approved):

zplizzi:
thanks! will merge in #283 so i can make the tests run 🙃

PR #273, opened 2026-08-10
Setup link redirects to login after claim

samuelren-imbue:
Redirect for any /setup link instead 403

### DIFF
--- a/compute_space/src/compute_space/web/app.py
+++ b/compute_space/src/compute_space/web/app.py
@@ -154,7 +154,7 @@ def _full_app_bootstrap(config: Config) -> None:
 def setup_already_done_get() -> Response[None]:
     """The claim link (``/setup?claim=...``) printed by ``openhost up`` should keep working
     after setup — send it to /login rather than a dead-end 403."""
-    return Redirect(path="/login")
+    return Redirect(path="/")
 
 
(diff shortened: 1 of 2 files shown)


compute_space/src/compute_space/web/app.py
```diff
@@ -149,8 +150,15 @@ def _full_app_bootstrap(config: Config) -> None:
     seed_first_boot(config)
 
 
-@route("/setup", http_method=[HttpMethod.GET, HttpMethod.POST], status_code=403, sync_to_thread=False)
-def setup_already_done() -> Response[str]:
+@get("/setup", sync_to_thread=False)
+def setup_already_done_get() -> Response[None]:
+    """The claim link (``/setup?claim=...``) printed by ``openhost up`` should keep working
+    after setup — send it to /login rather than a dead-end 403."""
+    return Redirect(path="/login")
```
zplizzi: i think it probably should just go to `/`, which will redirect to login if needed? maybe login redirects to / if already logged in but the former feels more correct.

zplizzi (approved):
lgtm after minor comment

compute_space/src/compute_space/web/app.py
samuelren-imbue: login does redirect to / but ill make it go directly to /

PR #275, opened 2026-08-11
diagnostics: speed up the diagnostics page (~4.5s → ~1s)

kilo-gi:
## Why

The owner-only Diagnostics bundle (`GET /api/diagnostics`) took ~4.5s to load. Measured on kilo-dev (15 running apps), the handler did nearly all its work **serially, on the event loop**:

| Phase | Cost | Cause |
|---|---|---|
| Per-app podman (`inspect` + `stats`, ×15, serial) | ~1.9s | 2 subprocess spawns per app, one app at a time |
| Reachability probes | ~0.95s | ran as a separate serial phase |
| `storage_status` | ~0.5s | walks app_data + `podman system df` |
| Health checks (15 serial) | ~0.24s | awaited one at a time |

Because the blocking subprocess calls ran on the single event loop, loading Diagnostics also froze the router for every proxied app for ~4.5s.

## What

Three changes to `collect_platform_diagnostics`, all confined to `core/diagnostics.py` (payload shape and `schema_version` unchanged):

- **Non-blocking** — every blocking probe (podman, `storage_status`) runs off the event loop via `asyncio.to_thread`; the independent parts (host facts, storage, reachability, per-app) are `asyncio.gather`ed, so the slowest bounds the wall-clock instead of the sum.
- **Parallel per-app** — the serial per-app loop becomes `asyncio.gather`, preserving per-app fault isolation and name-sorted order.
- **Batched podman** — one `podman ps` (authoritative running set) + one `podman stats` for the whole fleet instead of `inspect`+`stats` per app (~30 subprocess spawns → 2; **~1.9s → ~0.07s** measured on kilo-dev). The single-app bundle shares this path, so per-app resource collection has one implementation.

## Result

~4.5s → **~1s** (now bounded by the ~0.95s github reachability probe), and loading Diagnostics no longer stalls other apps on the box.

## Notes

An earlier revision also split the bundle into per-section endpoints so the page could paint fast sections first; that was dropped as too much surface for shaving perceived latency once the combined call was already ~1s. The net change here is just the core collector + its tests.

## Testing

Unit tests cover batched stats (one `ps` + one `stats`, no per-app `inspect`, regardless of app count), per-app concurrency, and the stats-parsing edge cases. Container-level behaviour is covered by the existing `--run-containers` suite. Perf numbers measured directly against kilo-dev.

### DIFF
--- a/compute_space/src/compute_space/core/diagnostics.py
+++ b/compute_space/src/compute_space/core/diagnostics.py
@@ -11,10 +11,6 @@ Two public entry points:
   - :func:`collect_app_diagnostics` — a per-app snapshot: the app's declared
     version + manifest git checkout, container status, plus a slim slice of the
     same host/system info so an app report is self-contained.
-
-Every collector is defensive: a failure gathering one field degrades that field
-to ``None``/an error string rather than failing the whole report, because a
-diagnostics bundle is most valuable precisely when the instance is unhealthy.
 """
 
 from __future__ import annotations
@@ -555,9 +551,12 @@ class _ContainerStatsBatch:
 
     Both collections are keyed by 12-char short id because ``podman stats``
     reports short ids while the DB stores full ids; callers truncate before
-    lookup. ``error`` is set when the podman probe itself failed — so every app
-    in the bundle surfaces the same reason instead of silently reading as
-    stopped.
+    lookup. ``error`` is set only when podman is *present but its probe failed*
+    (a real, unexpected fault) — so every app surfaces that reason instead of
+    silently reading as stopped. Podman merely being absent is not an error
+    here: it's reported once in ``container_runtime`` (see
+    :func:`_collect_container_runtime`), and every app then reads as having no
+    live stats, exactly like a stopped container.
     """
 
     running_short_ids: frozenset[str]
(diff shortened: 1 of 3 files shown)


kilo-gi:
@zplizzi I think this one could use a code review - it took me a few iterations to get the error handling down to something reasonable. There's still more isinstance() checks than I'd really like, but at least now they're all contained in the podman stats parsing code.

I did this because I noticed the diagnostics page was consistently taking >5s to load, and locking up other pages while it did. This change should make that a lot faster and keep it from locking up the router. This will also make surfacing the cpu+mem use in a more visual way easier+faster in https://github.com/imbue-openhost/openhost/pull/278

zplizzi:
will let @andrewlaack review this one, this is his code and i don't have strong opinions about it!

compute_space/src/compute_space/core/diagnostics.py
```diff
@@ -502,13 +506,133 @@ def _parse_stats_percent(value: object) -> float | None:
...
+    reports it authoritatively (see :func:`_collect_container_runtime`), so we
+    return an empty batch and let every app read as "no live stats" rather than
+    stamping the same message onto all of them. A podman that *is* present but
+    whose probe fails is an unexpected fault, and that we do surface in
+    ``error`` so apps don't masquerade as stopped.
+    """
+    if shutil.which("podman") is None:
+        return _ContainerStatsBatch(frozenset(), {})
+    try:
+        return _ContainerStatsBatch(_running_short_ids(), _stats_by_short_id())
```
andrewlaack: if the first succeeds and the second throws, the exception unwinds the whole expression and the podman ps answer is thrown away with it

compute_space/src/compute_space/core/diagnostics.py
```diff
@@ -735,59 +819,114 @@ async def _collect_app_summary(row: sqlite3.Row) -> AppDiagnosticsSummary:
...
 
+
+async def collect_platform_diagnostics(db: sqlite3.Connection, config: Config) -> PlatformDiagnostics:
+    """Assemble the full instance diagnostics bundle.
+
+    The independent parts are collected concurrently — the slowest (reachability)
+    bounds the wall-clock rather than the sum — and every blocking probe runs off
+    the event loop so serving diagnostics can't stall the router.
+    """
+    openhost_git, host_facts, storage, reachability, apps = await asyncio.gather(
```
andrewlaack, line 937: any unexpected raise in one member throws away the ones that already succeeded

andrewlaack (approved):

compute_space/src/compute_space/core/diagnostics.py
kilo-gi: ~Most~ All of these already have some reasonable error isolation underneath them. The main ways I can see that these would raise an error is if the database is inaccessible, or if we can't make basic calls like `platform.uname()`. In both cases we have bigger problems?

If you have a specific error/path you want to point out I'm happy to fix it, but I don't think trying to degrade more gracefully in general is worth the effort here.

compute_space/src/compute_space/core/diagnostics.py
kilo-gi: This seems realistic and easy enough, will push a fix shortly

PR #276, opened 2026-08-11
Add build_memory_mb limit for the app image build step

kilo-gi:
## What

Cap the real memory available to the app image **build** step (`podman build --memory`).

- Defaults to the app's runtime `memory_mb`, so behaviour is unchanged for existing apps.
- Optional `build_memory_mb` key under `[resources]` in `openhost.toml` overrides it independently — e.g. a big build that needs more RAM than the running container.
- Still allows for unlimited use of swap, so after https://github.com/imbue-openhost/openhost/pull/279 it shouldn't cause any builds to start failing, just slow down.

## Changes

- **manifest**: new `build_memory_mb: int | None` field + `effective_build_memory_mb` property that falls back to `memory_mb`.
- **`build_image`**: accepts `memory_mb` and appends `--memory=<n>m` (omitted when `None`).
- Both `build_image` callers (initial deploy + reload) pass `manifest.effective_build_memory_mb`. Each parses the manifest fresh from the repo.
- Docs (`manifest_spec.md`) + manifest/containers tests.

## Notes / decisions

- Named the key `build_memory_mb` for parallelism with the existing runtime key `memory_mb` (the issue said "something like build_memory_limit").
- Not persisted in the `apps` db row — it's a build-time-only value re-read from the manifest each build.
- No value validation; trusting podman's `--memory` to behave.

## Testing

Full `compute_space` suite: 965 passed, 43 skipped. New tests cover the `--memory` flag (present/absent) and the `memory_mb` fallback.

### DIFF
--- a/compute_space/src/compute_space/core/containers.py
+++ b/compute_space/src/compute_space/core/containers.py
@@ -192,7 +192,13 @@ def build_image(
         dockerfile_path,
     ]
     if memory_mb is not None:
+        # Cap build RAM at the app's declared limit for isolation, but allow
+        # unlimited swap (the host has a large swap file). A build that needs
+        # more than memory_mb then spills to swap instead of OOM-killing —
+        # keeping most builds working even with a small memory_mb, without
+        # letting a "small" app hog host RAM. --memory-swap=-1 requires --memory.
         cmd.append(f"--memory={memory_mb}m")
+        cmd.append("--memory-swap=-1")
     cmd.append(repo_path)
     logger.info("Building container image: %s", " ".join(cmd))
 
--- a/compute_space/src/compute_space/core/containers.py
+++ b/compute_space/src/compute_space/core/containers.py
@@ -174,8 +174,13 @@ def build_image(
     repo_path: str,
     dockerfile_rel_path: str,
     temp_data_dir: str | None = None,
+    memory_mb: int | None = None,
 ) -> str:
-    """Build the container image for an app.  Returns the image tag."""
+    """Build the container image for an app.  Returns the image tag.
+
+    ``memory_mb`` caps the memory available to the build via podman's
+    ``--memory`` flag; None leaves the build unconstrained.
+    """
     image_tag = f"openhost-{app_name}:latest"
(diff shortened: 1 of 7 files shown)


kilo-gi:
@zplizzi  you wanted to review this

compute_space/src/compute_space/core/containers.py
```diff
@@ -185,8 +190,16 @@ def build_image(
...
-        repo_path,
     ]
+    if memory_mb is not None:
+        # Cap build RAM at the app's declared limit for isolation, but allow
+        # unlimited swap (the host has a large swap file). A build that needs
+        # more than memory_mb then spills to swap instead of OOM-killing —
+        # keeping most builds working even with a small memory_mb, without
+        # letting a "small" app hog host RAM. --memory-swap=-1 requires --memory.
+        cmd.append(f"--memory={memory_mb}m")
+        cmd.append("--memory-swap=-1")
```
zplizzi, line 201: hmm i think builds shouldn't swap. i'd rather a build fail than be really slow and it not be clear why it's slow. and it's better for a build to request a big chunk of memory, and some other idle app get swapped out, than try to build an app in swap (which would do a lot of swapping since that memory is actively being used).

zplizzi: this is also more in line with your docs elsewhere.

kilo-gi: I'm 👍 on moving there, but right now even some of our core apps have a really small amount of memory allocated and their build uses more, so they would hard fail due to this change until they set their build memory flag. For example, we saw this with our test app.

Swap gives us a path out of that: we can make this change, do the update to set this for apps in our catalog, then make the enforcement strong.


So I guess I'm asking, is it ok if I fix this forward (which is a 2-line change) after we've done the catalog review and set build_memory_mb? 

Or I can revert, add the setting without enforcement for now and then readd enforcement after the catalog review.

zplizzi: yeah that seems fine, i guess you will need to test the catalog apps with a non-swapping instance tho to make sure they work.

PR #277, opened 2026-08-11
Add configurable swap file (default 16 GiB) with settings-page control

kilo-gi:
## Why

Several users have run into the memory limits of their instances, with apps getting OOM-killed under memory pressure. This provisions a **default 16 GiB swap file** on every host and lets the owner resize it from the settings page.

## Approach

Follows OpenHost's established system-config pattern — every host-level knob is configured in two mirrored places (ansible for fresh hosts, a `openhost_system_agent` migration for already-provisioned ones). The owner-facing resize goes through the root-run `openhost_system_agent` CLI rather than shelling out from the web app directly, so the privileged logic is shared with the migration and unit-tested.

**Source of truth is the swap file itself** (its on-disk size + `/etc/fstab` entry) — no new DB/config store; `get` reads live state.

## Changes

- **`openhost_system_agent/swap.py`** — single source of truth for swap path/default size and root-only create/resize/inspect logic. fallocate with a dd fallback, fstab persistence, `swapoff`+recreate on resize, `0` disables swap, bounds 0–256 GiB.
- **`SwapStatus`** protocol type + **`swap get` / `swap set <gib>`** CLI subcommands.
- **`ansible/tasks/swap.yml`** (create-if-absent) wired into `setup.yml`, `local_setup.yml`, `prebake.yml`; **`v0007_swap_file`** migration for existing hosts — non-destructive, so an owner-resized swap is never shrunk back to the default.
- **`GET`/`POST /api/settings/swap`** (owner-guarded, validated) + a **Swap** section in the settings page (`settings.html` / `settings.js`), mirroring the existing git-remote control.
- **Tests** (17): swap module logic (command sequence, dd fallback, create-only guard, fstab idempotency, root guard), the migration + a test asserting the ansible default stays in sync with `DEFAULT_SWAP_SIZE_GIB`, and the API handlers.

## Notes / decisions

- Swappiness left at the Ubuntu default (60 — already enough for the kernel to use swap under pressure). Happy to add a `vm.swappiness` sysctl knob if wanted.
- Swap path is `/swapfile`; `0` disables swap.

## Verification

- `openhost_system_agent` suite: 123 passed. compute_space settings/route/app suite: 311 passed.
- ruff + mypy clean (pre-commit).
- CLI `--help` confirms `swap get` / `swap set` parse.

### DIFF
--- a/ansible/local_setup.yml
+++ b/ansible/local_setup.yml
@@ -25,6 +25,7 @@
     - import_tasks: tasks/apt_packages.yml
     - import_tasks: tasks/containers.yml
     - import_tasks: tasks/journald.yml
+    - import_tasks: tasks/swap.yml
 
 - name: Dev environment
   hosts: localhost
--- a/ansible/prebake.yml
+++ b/ansible/prebake.yml
@@ -43,6 +43,7 @@
     - import_tasks: tasks/apt_packages.yml
     - import_tasks: tasks/containers.yml
     - import_tasks: tasks/journald.yml
+    - import_tasks: tasks/swap.yml
 
 - name: Dev environment
   hosts: all
--- a/ansible/setup.yml
+++ b/ansible/setup.yml
@@ -37,6 +37,7 @@
     - import_tasks: tasks/apt_packages.yml
     - import_tasks: tasks/containers.yml
     - import_tasks: tasks/journald.yml
+    - import_tasks: tasks/swap.yml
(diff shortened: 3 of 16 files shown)


kilo-gi:
Superseded by #279, which drops the settings-page control (resizing swap is an operator-level action done over SSH) and is rebased onto the latest main.

PR #278, opened 2026-08-11
diagnostics/system: split CPU/mem columns, sortable apps, resource donuts

kilo-gi:
Cosmetic changes to the Diagnostics and System Info pages. **Stacked on top of #275** (base branch is `kilo/sculptor/diagnostics-perf`) — no backend changes here.

## What

1. **Installed Apps table — split CPU/Memory + sortable** (`diagnostics.html`, `diagnostics.js`)
   - The combined `CPU / Mem` column is now two columns: **CPU** and **Memory**.
   - Every column header is clickable to sort; clicking again reverses direction, with a `▲`/`▼` indicator (plus `aria-sort`). Numeric columns sort numerically (not-running apps sort last); text columns sort case-insensitively with a stable name tie-break.
2. **Reachability moved** (`diagnostics.html`) — "Outbound Reachability" now sits **below** the Installed Apps table.
3. **System Info resource donuts** (`system.html`, `system.js`)
   - New **Resource Usage** section with CPU and Memory donut charts (same style as the existing "App data" storage donut — hover a slice for detail, plus a legend).
   - Both include an **Unused** headroom slice so overall box load is visible. Memory also breaks out a **System (non-app)** slice so the donut sums to total RAM. CPU capacity = `cpu_count × 100%`.
   - Refactored the existing storage donut into a reusable `donutHtml`/`wireDonut` pair shared by all three charts.

## Data source

Reuses the combined `/api/diagnostics` endpoint (owner-guarded) — the diagnostics table already renders from it, and the System Info charts pull `apps[].resources` (per-app `cpu_percent` / `memory_usage_bytes`) and `resource_pressure` (host totals) from the same payload. No Python changes.

## Testing

Rendered the real templates + CSS + JS through headless Chromium with mocked API data and verified: the split columns, click-to-sort (CPU descending puts blog 152% → analytics 65% → wiki 18% → not-running last), reachability placement, and both donuts. Verified CPU slices sum to `cpu_count × 100` and memory slices sum to total RAM.

### DIFF
--- a/compute_space/src/compute_space/web/templates/system.html
+++ b/compute_space/src/compute_space/web/templates/system.html
@@ -47,5 +47,6 @@ a snapshot from when the page loaded &mdash; reload to refresh.</p>
   "diagnosticsUrl": "/api/diagnostics"
 }
 </script>
+<script src="{{ static_url('js/chart.js') }}"></script>
 <script src="{{ static_url('js/system.js') }}"></script>
 {% endblock %}
--- a/compute_space/src/compute_space/web/templates/system.html
+++ b/compute_space/src/compute_space/web/templates/system.html
@@ -2,9 +2,9 @@
 {% block title_suffix %} - System Info{% endblock %}
 {% block content %}
 <h2 class="section-title">Resource Usage</h2>
-<p class="hint">CPU, memory, and app disk usage across running apps, with the box's
-remaining headroom shown as <em>Unused</em>. Hover a slice for details. These are
-a snapshot from when the page loaded &mdash; reload to refresh.</p>
+<p class="hint">Disk, CPU, memory, and swap usage for this box. CPU and memory are split
+across running apps, with the box's remaining headroom shown as <em>Unused</em>. Hover a
+slice for details. These are a snapshot from when the page loaded &mdash; reload to refresh.</p>
 <div class="usage-charts">
   <div class="usage-chart">
     <h3>Disk</h3>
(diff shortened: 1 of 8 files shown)


kilo-gi:
The new resource usage donuts:

image

The change to the diagnostics page, with split cpu/mem and sorting:

image

image

compute_space/src/compute_space/web/templates/system.html
```diff
@@ -1,6 +1,29 @@
 {% extends "layout.html" %}
 {% block title_suffix %} - System Info{% endblock %}
 {% block content %}
+<h2 class="section-title">Resource Usage</h2>
+<p class="hint">Disk, CPU, memory, and swap usage for this box. CPU and memory are split
```
zplizzi: nit: i'd be inclined to drop this line, i think this is pretty self explanatory? i prefer to keep things as minimal as possible in the UIs. or put it behind a little ? icon if you think it's important.

zplizzi (approved):
this looks great, i really like it! CPU feels weird to not be auto-updating and in a donut, maybe we'd wanna tweak that in the future, but i don't think it's worth putting more time into rn. ppl can always run htop in the terminal. eventually maybe a websocket-driven live table view of all the apps CPU/memory maybe makes sense or something.

kilo-gi:
The CI failure isn't due to my change afaict - looks like it's failing to download:


fatal: [136.64.67.76]: FAILED! => ***"changed": false, "dest": "/tmp/coredns-pinned.tgz", "elapsed": 0, "msg": "Connection failure: Remote end closed connection without response", "url": "https://github.com/coredns/coredns/releases/download/v1.14.4/coredns_1.14.4_linux_amd64.tgz"***

I guess I'll wait a bit and retry later?

compute_space/src/compute_space/web/templates/system.html
kilo-gi: Done

PR #279, opened 2026-08-11
Add configurable swap file (default 16 GiB)

kilo-gi:
## Why

Several users have run into the memory limits of their instances, with apps getting OOM-killed under memory pressure. This provisions a **default 16 GiB swap file** on every host so the kernel has somewhere to spill cold pages.

> Supersedes #277, which also added a settings-page control. Per feedback, resizing swap is an operator-level action — anyone who wants to change it can SSH in — so this PR drops the UI and keeps only the provisioning + an operator CLI.

## Approach

Follows OpenHost's established system-config pattern — every host-level knob is configured in two mirrored places (ansible for fresh hosts, an `openhost_system_agent` migration for already-provisioned ones). The privileged logic lives in the root-run `openhost_system_agent`.

**Source of truth is the swap file itself** (its on-disk size + `/etc/fstab` entry) — no new config store.

## Changes

- **`openhost_system_agent/swap.py`** — single source of truth for swap path/default size and root-only create/resize/inspect logic. fallocate with a dd fallback, fstab persistence, `swapoff`+recreate on resize, `0` disables swap, bounds 0–256 GiB.
- **`SwapStatus`** protocol type + **`swap get` / `swap set <gib>`** CLI subcommands, so an operator can resize over SSH: `sudo openhost_system_agent swap set 32`.
- **`ansible/tasks/swap.yml`** (create-if-absent) wired into `setup.yml`, `local_setup.yml`, `prebake.yml`; **`v0009_swap_file`** migration for existing hosts — non-destructive, so an operator-resized swap is never shrunk back to the default.
- **Tests**: swap module logic (command sequence, dd fallback, create-only guard, fstab idempotency, root guard), the migration + a test asserting the ansible default stays in sync with `DEFAULT_SWAP_SIZE_GIB`.

## Notes / decisions

- No settings-page UI (dropped from #277).
- Swappiness left at the Ubuntu default (60 — already enough for the kernel to use swap under pressure).
- Swap path is `/swapfile`; `0` disables swap.

## Verification

- `openhost_system_agent` suite: 137 passed.
- ruff + mypy clean (pre-commit).

### DIFF
--- a/ansible/local_setup.yml
+++ b/ansible/local_setup.yml
@@ -25,6 +25,7 @@
     - import_tasks: tasks/apt_packages.yml
     - import_tasks: tasks/containers.yml
     - import_tasks: tasks/journald.yml
+    - import_tasks: tasks/swap.yml
 
 - name: Dev environment
   hosts: localhost
--- a/ansible/prebake.yml
+++ b/ansible/prebake.yml
@@ -43,6 +43,7 @@
     - import_tasks: tasks/apt_packages.yml
     - import_tasks: tasks/containers.yml
     - import_tasks: tasks/journald.yml
+    - import_tasks: tasks/swap.yml
 
 - name: Dev environment
   hosts: all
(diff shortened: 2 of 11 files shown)


kilo-gi:
@zplizzi you wanted to review this

zplizzi:
lgtm, thanks!

PR #280, opened 2026-08-11
Add memory guard: warn on memory pressure and OOM kills

kilo-gi:
## Summary

Adds a **memory guard** daemon thread (`compute_space/core/memory_guard.py`), mirroring the existing storage guard and image pruner. Each tick (60s) it warns when:

- an app **nears its manifest memory limit** — two-threshold debounce (warn at 90%, re-arm below 80%) so a sustained condition warns once, not every tick;
- an app's container is **OOM-killed** — streamed per-app from `podman events --filter event=oom` (streaming, not polling `podman inspect`, since `--restart` resets `OOMKilled` to false and a poll would race the restart);
- the **host runs out of memory** and the kernel OOM killer reaps any process — read from the system journal via `journalctl --dmesg --follow --output=json`, catching machine-wide OOM that isn't scoped to any single app's cgroup.

Notifications are log lines with `TODO:` markers today; they become real notifications once that system exists.

## Design notes

- **One shared streamer for both detectors.** Both OOM sources are a subprocess emitting newline-delimited JSON (`journalctl` for the host, `podman events` per app), so the plumbing — non-blocking reads, partial-line buffering across ticks, reconnect-on-EOF — lives in one `_JsonEventStream`. The two readers are thin wrappers with their own line parsers.
- **Host reads the journal, not `/dev/kmsg`.** Reading the kernel log from the system journal means the host user needs only membership in the `systemd-journal` group, not the `CAP_SYSLOG` kernel capability — read-only access, no capability. The host parser is a *filter* (most kernel lines aren't OOM kills, so a non-match is normal).
- **Podman event parsing is total-and-loud.** Every line the OOM-filtered stream delivers *is* a kill, so `_parse_podman_event` commits to podman 5.8's top-level `ID`/`Name` keys (verified against the `v5.8.3` source and real output; podman is pinned `~=5.8`) and raises with the offending line on anything else, surfacing at the guard loop's `logger.exception` handler rather than silently dropping events.

## Granting journal access (additive drop-in)

The host user needs the `systemd-journal` group. Rather than edit the main unit or the shared `build_openhost_service_unit` builder (which lives in the `v0002` baseline migration), the group is granted through an **additive systemd drop-in** — so no existing migration is touched:

- New `ansible/files/openhost.service.d/10-journal-read.conf` carries `[Service]\nSupplementaryGroups=systemd-journal`; a task in `install_openhost_units.yml` installs it on fresh hosts.
- New **`v0010` system migration** writes the same drop-in + `daemon-reload` for existing hosts on their next self-update. The drop-in is kept **byte-identical** between the ansible copy and the migration constant (a test enforces it, mirroring the existing `RECLAIM_SCRIPT` pattern).
- The baseline unit and `build_openhost_service_unit` are unchanged; systemd merges the drop-in additively.

## Supporting changes

- Make `collect_app_resources` public (was `_collect_app_resources`) so the guard reuses it instead of duplicating the `podman stats` logic.
- Wire `ensure_memory_guard` into `_full_app_bootstrap`.

## Testing

- `test_memory_guard.py` (21 tests): pressure debounce, per-app OOM parsing (raising loudly on unrecognized shapes; skipping blank lines), streaming/reconnect, and host journal parsing (filtering non-OOM lines, journalctl-missing) — all mocked.
- `test_migration_journal_read_for_oom.py` (drop-in write + reload); `test_service_unit.py` pins the ansible-copy/migration-constant byte-identity.
- Full suites: `compute_space` 1232 passed / 43 skipped; `openhost_system_agent` 142 passed / 12 skipped. `mypy` + `ruff` clean.

## Notes for reviewers

- **Host OOM detection depends on journald reading kmsg (default) and the `systemd-journal` group grant.** No `@requires_containers`/integration test drives a real OOM through journalctl; the parsers are unit-tested against representative JSON. A smoke test on a real host would be a worthwhile follow-up.
- **No config gate**, unlike its sibling guards (`storage_min_free_mb`, `image_prune_interval_seconds`): the memory guard always runs and uses a process-singleton rather than the `db_path`-keyed pattern. Functionally equivalent in production (one `Config` per process), but flagged as an intentional inconsistency to confirm.

### DIFF
--- a/ansible/templates/openhost.service.j2
+++ b/ansible/templates/openhost.service.j2
@@ -22,14 +22,6 @@ StartLimitBurst=5
 [Service]
 Type=simple
 User=host
-# Add the host user to the systemd-journal group so it can read the system
-# journal. The memory guard tails the kernel log (via `journalctl --dmesg`) to
-# detect host-level (global) OOM kills — machine ran out of RAM and reaped some
-# process — which aren't scoped to any single app's cgroup and so aren't visible
-# via podman. Reading kernel messages otherwise needs root or CAP_SYSLOG; this
-# read-only group grants it without a kernel capability. Kept in sync with
-# build_openhost_service_unit in the system agent (a test asserts they agree).
-SupplementaryGroups=systemd-journal
 WorkingDirectory=/home/host/openhost
 Environment=PATH=/home/host/.pixi/bin:/home/host/openhost/.pixi/envs/default/bin:/usr/local/bin:/usr/bin:/bin
 Environment=OPENHOST_ROUTER_CONFIG=/home/host/.openhost/local_compute_space/config.toml
--- a/ansible/templates/openhost.service.j2
+++ b/ansible/templates/openhost.service.j2
@@ -22,6 +22,13 @@ StartLimitBurst=5
 [Service]
 Type=simple
 User=host
+# Grant CAP_SYSLOG so the (non-root) host user can read /dev/kmsg. The memory
+# guard tails the kernel log to detect host-level (global) OOM kills — machine
+# ran out of RAM and reaped some process — which aren't scoped to any single
+# app's cgroup and so aren't visible via podman. On a default dmesg_restrict=1
+# system reading the kernel log otherwise needs root; this ambient capability
+# lets User=host read it without giving up the rest of the sandbox.
+AmbientCapabilities=CAP_SYSLOG
 WorkingDirectory=/home/host/openhost
 Environment=PATH=/home/host/.pixi/bin:/home/host/openhost/.pixi/envs/default/bin:/usr/local/bin:/usr/bin:/bin
 Environment=OPENHOST_ROUTER_CONFIG=/home/host/.openhost/local_compute_space/config.toml
--- a/ansible/templates/openhost.service.j2
+++ b/ansible/templates/openhost.service.j2
@@ -22,13 +22,14 @@ StartLimitBurst=5
 [Service]
 Type=simple
 User=host
-# Grant CAP_SYSLOG so the (non-root) host user can read /dev/kmsg. The memory
-# guard tails the kernel log to detect host-level (global) OOM kills — machine
-# ran out of RAM and reaped some process — which aren't scoped to any single
-# app's cgroup and so aren't visible via podman. On a default dmesg_restrict=1
-# system reading the kernel log otherwise needs root; this ambient capability
-# lets User=host read it without giving up the rest of the sandbox.
-AmbientCapabilities=CAP_SYSLOG
+# Add the host user to the systemd-journal group so it can read the system
+# journal. The memory guard tails the kernel log (via `journalctl --dmesg`) to
+# detect host-level (global) OOM kills — machine ran out of RAM and reaped some
+# process — which aren't scoped to any single app's cgroup and so aren't visible
+# via podman. Reading kernel messages otherwise needs root or CAP_SYSLOG; this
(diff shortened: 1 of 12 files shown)


ansible/templates/openhost.service.j2
```diff
@@ -22,6 +22,13 @@ StartLimitBurst=5
 [Service]
 Type=simple
 User=host
+# Grant CAP_SYSLOG so the (non-root) host user can read /dev/kmsg. The memory
+# guard tails the kernel log to detect host-level (global) OOM kills — machine
+# ran out of RAM and reaped some process — which aren't scoped to any single
+# app's cgroup and so aren't visible via podman. On a default dmesg_restrict=1
+# system reading the kernel log otherwise needs root; this ambient capability
+# lets User=host read it without giving up the rest of the sandbox.
+AmbientCapabilities=CAP_SYSLOG
```
zplizzi: this is maybe insecure?

> Kernel symbol addresses defeat KASLR, which is the main mitigation standing between a memory-corruption bug and a working exploit. A process with CAP_SYSLOG turns "attacker needs an infoleak plus a write primitive" into "attacker needs a write primitive."

claude suggests some alternatives: https://claude.ai/share/eb3d5f2a-9ff0-427e-883f-b50243bcfbf1

openhost_system_agent/src/openhost_system_agent/migrations/versions/v0002_baseline.py
```diff
@@ -72,6 +72,15 @@ def build_openhost_service_unit(host_uid: int) -> str:
...
         "User=host\n"
+        # Grant CAP_SYSLOG so the (non-root) host user can read /dev/kmsg. The
+        # memory guard tails the kernel log to detect host-level (global) OOM
+        # kills — machine ran out of RAM and reaped some process — which aren't
+        # scoped to any single app's cgroup and so aren't visible via podman. On
+        # a default dmesg_restrict=1 system reading the kernel log otherwise
+        # needs root; this ambient capability lets User=host read it without
+        # giving up the rest of the sandbox. Kept in sync with the ansible
+        # template.
+        "AmbientCapabilities=CAP_SYSLOG\n"
```
zplizzi: see above comment, but wouldn't we need a real migration to add this for existing systems?

zplizzi (changes requested):
high level i like this. thought: if we catch an OOM kill for an app, should we mark that app as errored with an appropriate error message (and not restart the app)? that would be even clearer than just a notification; you might not notice a notification if you already had a bunch of unreads. but on the flip side, maybe you'd prefer some apps auto-recover themselves. i don't feel strongly either way (but i don't think we should make it configurable; we should just pick one or the other)

i would like to re-review after the CAP_SYSLOG changes/response.

ansible/templates/openhost.service.j2
kilo-gi: kk, moved this over to adding the systemd-journal group and reading from there instead.

kilo-gi:
I think I'm against forcing apps down after an OOM kill - often an OOM kill 'just works' and resets a memory leak or similar. Plus I don't think a lot of people are going to have actual monitoring on their instance, so any externally-visible app that gets OOM killed just... goes down for the people using it?

That's also not how most other services work. I am 👍 on us considering a host tainted and 'you should really reboot' after any OOM killer run, but I think we're expecting the mode case here to be that some app hit their memory limit, not 'OOM killer ran an could have hit any of 100 load-bearing processes on your box, it's impossible to tell if it's still working right or not'

zplizzi (approved):

PR #282, opened 2026-08-11
Make save remote also apply updates

samuelren-imbue:


andrewlaack:
Not sure we want this given #270 setRemote already surfaces the pin as an available update, and the Update button now streams live progress on /updating. Applying unconditionally also restarts (and reinstalls) even when the pin already resolves to HEAD

PR #284, opened 2026-08-12
Annotate routes

samuelren-imbue:
Add exception types in raises=[], will be used for openAPI docs


compute_space/src/compute_space/web/routes/api/system.py
```diff
@@ -69,16 +75,16 @@ class CreatedToken:
...
-class ErrorResponse:
-    error: str
-
-
 @attr.s(auto_attribs=True, frozen=True)
 class OkResponse:
     ok: bool
 
 
+def _system_api_error_response(request: Request[Any, Any, Any], exc: HTTPException) -> Response[Any]:
```
zplizzi: how do the litestar exceptions get formatted by default, without a special handler? they should already get serialized to json

samuelren-imbue: litestar serializes with: {"status_code": 400, "detail": "...", "extra": ..., "headers": ...} but our current js looks for whether data.error exists. the extra stuff in litestar doesn't need to get passed, so I thought it best to not.

compute_space/src/compute_space/web/routes/api/apps.py
```diff
@@ -276,28 +257,36 @@ async def _pin_refless_to_landed_branch(repo_url: str | None, repo_path: str) ->
 # ─── routes ────────────────────────────────────────────────────────────────
 
 
-@post("/api/clone_and_get_app_info", status_code=200, guards=[require_owner_auth])
+@post(
+    "/api/clone_and_get_app_info",
+    status_code=200,
+    guards=[require_owner_auth],
+    raises=[ValidationException, NotAuthorizedException, InternalServerException],
```
zplizzi: i would probably drop `InternalServerException` in the annotation on these routes. i think the annotatations/docs should only include things that the user can actually do something about. if it's a state that should never be possible to reach in normal operation, i think we don't need to annotate/document it.

zplizzi: you can still raise it though, it's more of just like an assert failure and kinda implied that a client ought to be able to handle these types of failures with some default fail handler.

compute_space/src/compute_space/web/routes/api/apps.py
```diff
@@ -506,55 +496,58 @@ def _app_diagnostics_filename(app_name: str) -> str:
...
-async def stop_app(
-    app_id: FromPath[str], db: NamedDependency[sqlite3.Connection]
-) -> Response[OkResponse] | Response[ErrorResponse]:
-    app_row, err = _resolve_app_or_error(app_id, db)
-    if err is not None:
-        return err
-    assert app_row is not None
+@post(
+    "/stop_app/{app_id:str}",
+    status_code=200,
```
zplizzi, line 543: reminder - drop these status_code=200 if possible

compute_space/src/compute_space/web/routes/api/apps.py
```diff
@@ -889,18 +883,15 @@ async def remove_app(
...
-                        "S3-side bytes wouldn't actually be deleted.  "
-                        "Either restore the archive mount and retry, or "
-                        "use keep_data=1 to remove the app while leaving "
-                        "its data in place."
-                    )
-                ),
-                status_code=503,
+            raise ServiceUnavailableException(
+                detail=(
+                    "Archive backend is not healthy; refusing to "
```
zplizzi: nit: reflow to 120 chars wide

compute_space/src/compute_space/web/routes/api/apps.py
zplizzi: and on other instances of this

compute_space/src/compute_space/web/routes/docs.py
```diff
@@ -865,35 +866,28 @@ def docs_slug(slug: FromPath[str]) -> Response[str]:
     return _render_doc(slug)
 
 
-def _missing_src_response() -> Response[str] | None:
-    """The 503 for a broken checkout, or ``None`` when ``docs/src/`` is present."""
+def _require_docs_src() -> None:
+    """Raise the 503 for a broken checkout; return normally when ``docs/src/`` is present."""
```
zplizzi: maybe worth annotating in docstring what these helpers raise, esp the 2 after this which may not be obvious

compute_space_cli/src/compute_space_cli/helpers.py
```diff
@@ -22,6 +22,21 @@ def resolve_app_id_by_name(url: str, token: str, app_name: str) -> str:
     return app_id
 
 
+def error_message(body: object, fallback: str) -> str:
```
zplizzi: this looks a lil messy but i haven't thought about it, just flagging if there's any easy cleanups you see here.

zplizzi (approved):
minor comments but looks great overall! seems definitely cleaner than before, thanks for doing this.

compute_space/src/compute_space/web/routes/api/apps.py
samuelren-imbue: some of them have actual info (system.py gives info on docker build cache) so ill keep some

PR #290, opened 2026-08-13
Stream app logs over a WebSocket (shared process-follow primitive)

kilo-gi:
Stacked on #280 — base is that branch, so this PR's diff is only the log-streaming work.

## What & why
The app-detail page re-fetched the **entire** container log every 1–3s and rewrote the whole `<pre>`, so a large log (bifrost's ran to several MB) made the page lag badly. This replaces the poll with a live WebSocket stream, and extracts the "follow a subprocess for log lines" plumbing so the memory guard shares it.

## Changes
- **`core/process_stream.py`** (new): `stream_process_lines(argv)` — spawns a command and yields stdout line-by-line until EOF, terminating/reaping the child on early exit or shutdown. Errors aren't papered over: EOF is the only normal terminator; a missing binary / read error propagates for the caller to classify.
- **`core/log_stream.py`** (new): build-log tail (Python file-existence guard so a permission error fails loud instead of `tail`'s silent retry) followed by the live container log via `podman logs --follow` through the shared helper.
- **`GET /app_logs_stream/{app_id}`** (WebSocket): inline owner-auth, races client-disconnect + shutdown to reap the follow. The detail page no longer reads the whole log up front.
- **`app-detail.js`**: consume over a WebSocket, appending batched line(s) once per animation frame (bounded + front-trimmed) so the ~2000-line replay renders in linear time instead of O(n²) layout thrash. Reconnect re-primes so the replayed tail replaces (not duplicates) the visible log.
- **`core/memory_guard.py`**: converted onto the same shared helper — its `journalctl` and `podman events` readers are now reconnecting async followers, and the guard runs its own asyncio loop on its daemon thread (separate from the request loop, so its blocking `podman stats`/sqlite work can't stall requests).
- Lower the container-log `max-size` safety net from 10mb to 1mb.

## Tests
New unit coverage for `process_stream`, `log_stream`, the WS route, and the reworked memory guard (44 tests); full `compute_space` suite green (1255 passed / 43 skipped), ruff + mypy clean.

### DIFF
--- a/compute_space/src/compute_space/core/log_stream.py
+++ b/compute_space/src/compute_space/core/log_stream.py
@@ -129,12 +129,17 @@ async def stream_app_logs(app_id: str, build_log_path: str) -> AsyncIterator[str
     The app's ``(status, container_id)`` is re-queried while a build is in flight so we
     can hand off from tailing ``docker.log`` to following the container as it comes up.
     """
+    emitted = False
     if _still_building(app_id):
         # Build in flight: follow docker.log live until the container comes up.
         async for line in _follow_build_until_container(app_id, build_log_path):
+            emitted = True
             yield line
-    elif os.path.exists(build_log_path):
-        # Already built: replay the build log's tail once, then show container logs.
+
+    # Replay the build log's tail once if the live follow produced nothing — either the
+    # app was already built, or the build ended before docker.log appeared and it only
+    # landed just now. Guarded on ``emitted`` so a followed build isn't replayed twice.
+    if not emitted and os.path.exists(build_log_path):
         async for raw in stream_process_lines(
             _tail_argv(build_log_path, follow=False), merge_stderr=True, raise_on_error=True
         ):
--- a/compute_space/src/compute_space/core/log_stream.py
+++ b/compute_space/src/compute_space/core/log_stream.py
@@ -5,26 +5,20 @@ chunks (each with no trailing newline) that a transport layer — currently the
 WebSocket route in ``web/routes/api/apps.py`` — forwards to the browser, which
 appends each chunk as one line.
 
-Two sources, matching the two phases of an app's life:
-
-* The **build log** (``docker.log``) is a file our own build process appends to
-  (only ever rotated on an explicit reload, which reloads the whole page), so we
-  read it directly by byte offset. A missing file is the *expected* "build hasn't
-  written yet" state and reads empty; any other read error (e.g. permission
-  denied) propagates rather than being turned into a silent empty log.
-* The **container log** comes from ``podman logs --follow`` via the shared
-  :func:`compute_space.core.process_stream.stream_process_lines` primitive (the
-  same one the memory guard uses for its ``journalctl`` / ``podman events``
-  streams). podman — not us — owns incrementalism and log rotation, so we never
-  compute an offset into a file that can move underneath us. We just strip its
-  ``--timestamps`` prefix and ANSI codes, reproducing what the old one-shot
-  ``podman logs`` call showed.
+Both sources are followed through the same shared primitive,
+:func:`compute_space.core.process_stream.stream_process_lines` (also used by the
+memory guard): the container log via ``podman logs --follow``, and the build log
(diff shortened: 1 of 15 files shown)


zplizzi (approved):
code-owner parts look fine, will bump to @andrewlaack to review the rest. the high level idea of streaming logs over websockets was discussed and seems good to me.

compute_space/src/compute_space/core/memory_guard.py
```diff
@@ -215,56 +165,71 @@ def _parse_podman_event(line: bytes) -> _ContainerOomKill:
...
-        finally:
-            db.close()
-
-        self._report_container_ooms(container_ooms, rows)
-        for row in rows:
+        ``_parse_podman_event`` raises on an unrecognized shape rather than silently
+        dropping the kill; that surfaces at ``run``'s caller (the loop's handler).
+        """
+        async for line in _follow_lines(_PODMAN_OOM_EVENTS, "Per-app OOM detection"):
+            kill = _parse_podman_event(line)
```
andrewlaack: This ValueError escapes into run()'s gather, which keeps the default return_exceptions=False, so one unparseable event line tears down the two sibling detectors and restarts the whole guard 5s later. That blinds the unrelated host journal follower and the pressure loop, where the old code only lost a tick of one detector. A catch per line should fix this.

compute_space/src/compute_space/core/process_stream.py
```diff
@@ -0,0 +1,131 @@
...
+        while line := await proc.stdout.readline():
+            yield line.rstrip(b"\n")
+        if raise_on_error:
+            await proc.wait()
+            if proc.returncode:
+                raise RuntimeError(f"`{argv[0]}` exited with status {proc.returncode}")
+    finally:
+        _active.discard(proc)
+        _terminate(proc)
+        with contextlib.suppress(Exception):
```
andrewlaack, line 121: CancelledError is a BaseException, so suppress(Exception) does not stop it here. A cancellation that arrives while this is suspended on proc.wait() escapes, and the child is then signalled but never reaped, which breaks the guarantee the module docstring makes. suppress(BaseException) is the wrong fix because it swallows cancellation, whereas a shielded or bounded reap keeps both properties.

compute_space/src/compute_space/web/static/js/app-detail.js
```diff
@@ -193,22 +187,92 @@ function clearCacheAndReload() {
...
+    }
+
+    // Tracked so a reset can tell the old socket's onclose not to reconnect.
+    var currentWs = null;
+
+    function startLogStream() {
+        var proto = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
+        var ws = new WebSocket(proto + '//' + window.location.host + config.appLogsStreamUrl);
+        currentWs = ws;
+        ws.onmessage = function(e) {
```
andrewlaack, line 266: onclose has the ws !== currentWs guard but onmessage does not. After resetLogStream(), a message still queued on the superseded socket consumes the logPrimed reset, so the new socket's replayed tail appends instead of replacing and the pane ends up showing both containers. Adding the same guard here fixes it.

compute_space/src/compute_space/core/log_stream.py
```diff
@@ -0,0 +1,147 @@
...
+    """Yield an app's logs as text chunks: build-log tail, then live container logs.
+
+    The app's ``(status, container_id)`` is re-queried while a build is in flight so we
+    can hand off from tailing ``docker.log`` to following the container as it comes up.
+    """
+    if _still_building(app_id):
+        # Build in flight: follow docker.log live until the container comes up.
+        async for line in _follow_build_until_container(app_id, build_log_path):
+            yield line
+    elif os.path.exists(build_log_path):
```
andrewlaack: This is an elif, so when the build ends before docker.log ever appears, _follow_build_until_container yields nothing and this replay is skipped even if the file lands a moment later. The viewer is left with an empty pane. Making it a standalone if, guarded on having emitted nothing, covers that case.

andrewlaack (commented):
A few comments.

andrewlaack (approved):

PR #296, opened 2026-08-17
Fix loguru log calls that used %-style placeholders

andrewlaack:
Several modules import the loguru logger from compute_space.core.logging but log with stdlib %-style placeholders (%s, %d, %r, %.0f) plus trailing args. loguru formats with str.format ({}-style), so it never substituted them: the raw template was emitted with a literal %s/%d and the args were dropped. This is visible in oh logs, e.g. "Container build attempt %d/%d for %s failed", "Running container: %s", "default_apps: %s deployed", and the cpu_millicores deprecation warning ("App '%s' uses deprecated 'cpu_millicores'").

This converts those message templates to loguru {}-style ({}, {!r}, {:.0f}) across the affected modules (88 call sites, including logger.opt(...).warning chained calls). Only message templates change; trailing args are unchanged, and stdlib logging interception (record.getMessage) is unaffected.

Validated: py_compile of all changed files; an AST check that no logger call retains a %-placeholder and that every template's {}-field count matches its positional-arg count; and a loguru replay of all converted templates.


zplizzi:
lgtm

zplizzi (approved):

PR #297, opened 2026-08-18
Router: fix authenticated app form POSTs (Origin: null) + 403 for unauth unsafe methods

kilo-gi:
## Summary

Fixes the **"form POST bounced to `/login` → 405"** report — every state-changing form action in a routed app (add feed, import, refresh, delete, save settings…) failed even with a valid owner session. Two related router-side defects:

1. **Root cause (auth):** `verify_owner_auth` rejected an authenticated same-origin POST that carries **`Origin: null`**. A real browser sends `Origin: null` for some legitimate same-origin top-level form POSTs (see the breakdown below) while the request stays same-origin and still sends the `SameSite=Lax` session cookie. The strict Origin-match treated `null` as cross-origin → rejected.
2. **Amplifier (lossy failure response):** that rejection was answered with a `302 → /login` **regardless of method**. A browser follows the 302 by re-issuing the request as a bodyless **GET**, so the POST-only app route returns **405** and the action is silently lost.

### Smallest thing this broke

An utterly ordinary same-origin POST form — the *only* special ingredient is a **no-referrer policy on the page**. It does **not** go on the `<form>` (`referrerpolicy` isn't a valid `<form>` attribute); it's set document-wide via `<meta>`, or — as Miniflux does — via a `Referrer-Policy: no-referrer` **response header**.

```html
<!-- served by any app at https://<app>.<zone>/… -->
<meta name="referrer" content="no-referrer">
<form action="/save" method="post">
  <button>Save</button>
</form>
```

Before the fix, clicking **Save** → same-origin `POST /save` with `Origin: null` → router bounced to `/login` → browser re-issued as GET → **405**. No JS, no iframe, no cross-origin anything.

## Fixes

**1. Accept `Origin: null` only when Fetch-Metadata corroborates same-origin** (`web/auth/auth.py`, `is_user_request_same_origin`). `Sec-Fetch-Site` is set by the browser from the true initiator and is a **forbidden header name** (JS cannot forge it). A genuine app-posting-to-itself reports `same-origin`; a cross-app request reports `same-site`/`cross-site`. So we honor `same-origin` and keep everything else rejected — unblocking legitimate app forms **without reopening cross-app CSRF**. Browsers that omit `Sec-Fetch-Site` keep failing closed on `null`. The `/logout` CSRF guard (`verify_same_origin`) is unchanged.

Behavior matrix (authenticated owner POST to `miniflux.<zone>`):

| Origin | Sec-Fetch-Site | Result |
|---|---|---|
| `null` | `same-origin` | ✅ authorized (the fix) |
| `null` | `same-site` / `cross-site` | ❌ rejected |
| `null` | *(absent)* | ❌ rejected (fail closed) |
| `https://other.<zone>` | `same-origin` (spoof) | ❌ rejected |
| `https://miniflux.<zone>` | — | ✅ authorized |
| *(absent)* | — | ✅ authorized (normal nav) |

<details>
<summary><b>How a browser ends up sending <code>Origin: null</code> (and why the gate is safe for each)</b></summary>

There are two fundamentally different mechanisms, with opposite security properties:

**A — the `Origin` *header value* is nulled, but the document's origin is real.** The page is a normal same-origin document; only the header is serialized as `null`, so the `SameSite=Lax` cookie **is** sent and `Sec-Fetch-Site` reflects the real origins.
- `Referrer-Policy: no-referrer` response header on the app — **the Miniflux case.** This is the only referrer policy that nulls the Origin on a same-origin, same-scheme POST.
- `<meta name="referrer" content="no-referrer">` in the page.
- `rel="noreferrer"` on the submitting element.
- HTTPS→HTTP downgrade under `strict-origin*` / `no-referrer-when-downgrade` (only on scheme downgrade — N/A when everything is HTTPS).

**B — the document's *origin itself* is opaque (null).** The whole context is opaque, so it is treated as **cross-site for cookies** (no Lax cookie) and `Sec-Fetch-Site` is `cross-site`.
- Sandboxed iframe (`<iframe sandbox>` without `allow-same-origin`).
- `Content-Security-Policy: sandbox` on the document.
- Sandboxed `srcdoc` iframe.
- `data:` URL document; `file://` document (browser-dependent).

**Plus:** a cross-origin redirect in the request chain nulls the Origin on later hops (tainted-origin; mostly a CORS/`fetch` concern, not top-level form nav).

| Cause | Doc origin | Lax cookie sent? | `Sec-Fetch-Site` | Gate result |
|---|---|---|---|---|
| A: `no-referrer` etc., **app posts to itself** | real (same-origin) | ✅ yes | `same-origin` | ✅ **accepted** (the legit case) |
| A: `no-referrer`, **app-A → app-B top-level form** | real (app-A) | ✅ yes (same-site) | `same-site` | ❌ rejected (the CSRF we must block) |
| B: sandbox / `data:` / `file:` | opaque | ❌ no | `cross-site` | ❌ rejected (fails auth *and* the gate) |
| cross-origin redirect | varies | typically no | `cross-site` | ❌ rejected |

The load-bearing rows are the first two: both send `Origin: null` **with** the owner's cookie, and the only thing distinguishing "app posting to itself" from "app-A forging to app-B" is `Sec-Fetch-Site` (`same-origin` vs `same-site`). That is exactly why the gate keys on `same-origin` rather than accepting `null` unconditionally. Mechanism B never reaches the authenticated branch (no Lax cookie), so it is doubly rejected.

</details>

**2. Return 403 (not 302→/login) for unauthenticated unsafe methods** (`web/auth/auth.py` `auth_required_response`, wired into `SubdomainProxyMiddleware` and the `NotAuthorizedException` handler in `web/app.py`). When owner auth *does* legitimately fail on a POST/PUT/PATCH/DELETE, answer honestly with 403 so the method/body are never silently downgraded to a GET; only navigational GET/HEAD get the `/login` redirect.

## Tests

- `tests/test_verify_owner_auth_origin.py` — the `Origin: null` matrix above (real bug case + CSRF non-regression).
- `tests/test_auth_required_response.py` — GET/HEAD → 302, POST/PUT/PATCH/DELETE → 403.
- `tests/test_subdomain_proxy_auth_redirect.py` — real `SubdomainProxyMiddleware`: unauth POST → 403 (not proxied); GET → 302 `/login`.

Full suite: **1280 passed, 43 skipped**; ruff + mypy clean. (Rebased onto current `main`, domain-in-DB model — no config changes.)

## Verified live on kilo-dev ✅

Deployed on the live instance (`main` + this auth diff) and re-run firsthand in an authenticated owner browser:

- **Refresh all feeds** — `POST /feeds/refresh` → `302 → /feeds` (success; **no** `/login` bounce). True request headers: `Origin: null`, `Sec-Fetch-Site: same-origin`, `Sec-Fetch-Mode: navigate` — the gate condition, exactly as predicted.
- **Add a feed** — `POST /subscribe` → `302 → /feed/1/entries` (feed added, 30 entries fetched). Feeds list shows "Feeds (1)".

So the authenticated `Origin: null` POST is authorized and the form submits — not merely a 405→403 change; the actual POST goes through.

### DIFF
--- a/compute_space/src/compute_space/tests/_litestar_helpers.py
+++ b/compute_space/src/compute_space/tests/_litestar_helpers.py
@@ -27,6 +27,45 @@ from compute_space.db import provide_db
 from compute_space.web.helpers.zone import ZONE_SCOPE_KEY
 
 
+def make_http_scope(
+    method: str,
+    path: str,
+    *,
+    host: str,
+    headers: dict[str, str] | None = None,
+    cookie: str | None = None,
+    client: tuple[str, int] | None = None,
+    extra_scope: dict[str, Any] | None = None,
+) -> dict[str, Any]:
+    """Build a minimal HTTP ASGI scope for auth / middleware unit tests.
+
+    Callers wrap the result however they need — ``Request(scope)``, ``ASGIConnection(scope)``, or pass
+    the raw dict straight into an ASGI middleware.  ``cookie`` (a raw ``name=value`` string) and
+    ``headers`` become request headers; ``extra_scope`` merges in extra scope keys (e.g. the
+    ``ZONE_SCOPE_KEY`` the proxy middleware would normally stash).
+    """
+    raw_headers: list[tuple[bytes, bytes]] = [(b"host", host.encode())]
+    if cookie is not None:
+        raw_headers.append((b"cookie", cookie.encode()))
+    for name, value in (headers or {}).items():
+        raw_headers.append((name.lower().encode(), value.encode()))
+    scope: dict[str, Any] = {
+        "type": "http",
+        "method": method,
+        "path": path,
+        "raw_path": path.encode(),
+        "query_string": b"",
+        "scheme": "http",
+        "server": ("127.0.0.1", 8080),
+        "root_path": "",
+        "headers": raw_headers,
+        **(extra_scope or {}),
+    }
+    if client is not None:
+        scope["client"] = client
+    return scope
+
+
 def stash_zone_middleware(app: ASGIApp) -> ASGIApp:
     """Test stand-in for the zone-stashing half of ``SubdomainProxyMiddleware``: put the DB primary
     in the request scope so ``zone_for_request`` resolves in minimal test apps that omit the full
--- a/compute_space/src/compute_space/tests/test_auth_required_response.py
+++ b/compute_space/src/compute_space/tests/test_auth_required_response.py
@@ -15,6 +15,7 @@ import pytest
 from litestar import Request
 
 from compute_space.core.domains import Domain
+from compute_space.tests._litestar_helpers import make_http_scope
 from compute_space.tests.conftest import _make_test_config
 from compute_space.web.auth.auth import auth_required_response
 from compute_space.web.auth.auth import is_login_redirectable_method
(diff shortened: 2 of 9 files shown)


zplizzi:
FYI i want to think about this in the context of some other related changes I wanted to make; probs won't get to it immediately.

PR #298, opened 2026-08-18
Add a design system to the compute space web UI and make it responsive

zplizzi:
Two commits, reviewable in order:

1. **Add a design-token/component layer and make the web UI responsive** — the
   foundation plus the dashboard, built to the Figma design
   ([Product Branding, node 1689-5150](https://www.figma.com/design/0iHhQDfxxoNo6twrpKIRxe/Product-Branding?node-id=1689-5150)).
2. **Roll the design system out across the rest of the compute space** — extends
   the vocabulary to the components the Figma doesn't cover, and converts every
   remaining page.

## Approach

No JS framework and no build step. Components are Jinja macros whose identity is
a **CSS class contract**, so whichever way the JS question goes later
(Solid / Lit / plain custom elements), it's a template change rather than a
rewrite — the tokens, stylesheet and markup carry over intact.

```
static/css/tokens.css        palette, mono type scale, spacing, shadow, widths
static/css/base.css          reset, sky/grass page frame, typography, tables, forms
static/css/components.css    the style contract for each macro
static/js/dom.js             el/replace/badge builders, replacing innerHTML concat
templates/_components/       icon, button, card, section_head, app_row, spec_list
templates/_layout_narrow.html  shared shell for login / setup / identity-approve
```

`layout.html`'s 60-line inline `<style>` is gone, along with the per-template
`<style>` blocks and most scattered `style="..."` attributes. `login`, `setup`
and `identity_approve` previously duplicated their own `<head>` three times.

## Extending the design language beyond the comps

There are comps for one screen, so for everything else I derived components from
what the Figma *does* fix rather than inventing a parallel style: monospace;
sharp corners (radius 0 everywhere it matters); 1px `#dadada` borders on
`#fcfcfc`; a hard `2px 2px 0` unblurred shadow as the only "raised" signal;
uppercase + `0.72px` tracking as the "this is a label" signal; and flat pastel
fills used **only** to carry meaning.

| Component | Derivation |
|---|---|
| `.badge` | The "+ NEW" button is already a flat pastel square with tracked uppercase text — the badge is that, generalised to ok/info/warn/error. |
| `.notice` | The same tints at panel scale, with a 3px bar carrying status so the message stays plain text. |
| `.meter` | Segmented, not smooth — the blocky fill echoes the pixel-art clouds and grass rather than a modern progress bar. |
| Tables | Adopt the app list's vocabulary: horizontal rules only, columns carried by mono alignment, uppercase tracked `th`. |
| Checkbox / radio / select | Natives arrive rounded and OS-tinted, fighting the sharp flat treatment; restyled square. |

Kept as-is: the **donut charts**. A donut genuinely beats a bar for the
multi-segment storage breakdown, where proportion *between parts* is the point —
its palette moved onto the brand pastels instead. The single-value CPU/memory/
swap charts would read better as `.meter`s, but that's a behaviour change rather
than a restyle.

## Mobile

The app was desktop-only: `@media` appeared **zero** times, and only
`updating.html` had a viewport meta — every other page rendered zoomed out on a
phone.

- Viewport meta in both layouts
- The 21 `<table>`s split by kind: key/value tables become `.spec-list`
  (collapses via a container query), `form-table` stacks below 640px which makes
  **settings usable on a phone for the first time**, and genuinely tabular data
  scrolls horizontally instead of forcing the page wide
- Cards reflow with `auto-fit`; 36px touch targets on coarse pointers
- The 7 nav tabs fit the content column, scroll on narrow, active tab scrolled
  into view

## Bugs found along the way

1. **`[hidden]` did nothing on app rows.** A class-level `display: flex` outranks
   the UA stylesheet's `[hidden] { display: none }`, so the dashboard's row
   hiding silently failed. Fixed globally in `base.css`.
2. **XSS in `tokens.js`** — a user-controlled token name interpolated straight
   into `innerHTML`. There were three near-identical hand-rolled HTML escapers
   in the codebase precisely because every interpolation was a place to forget
   one; `dom.js` removes the class of bug by putting text in as text.
3. **App names never truncated** — `text-overflow: ellipsis` can't fire on a flex
   item without `min-width: 0`.
4. **`section_head` emitted a `<span>`**, costing every converted page its
   heading outline. Caught by an existing test; fixed the component, not the test.
5. **`list_permissions_v2`** used the deprecated inferred query-param style —
   the one deprecation warning left in the test run.

## /design

A gallery rendering every component in every state from the same macros the real
pages use, so drift shows up in one place. It doubles as a screenshot target for
responsive regressions, and it caught bug #3 immediately via a deliberately
over-long demo app name.

## Notes

- Dashboard rows are name + status + Details, per the design; all app actions
  live on the detail page.
- I did **not** rewrite the page scripts wholesale — ~1,700 lines of working,
  untested JS is a poor thing to churn for its own sake. I converted the sites
  that emit styled markup (which had to change anyway) and left the logic alone.
- `update-progress.css` deliberately still uses literal values, not tokens: the
  updater inlines it into a standalone page served *while compute_space is down*,
  so neither `tokens.css` nor the webfont is reachable. Commented in place.
- The font loads from Google Fonts, a second external dependency alongside the
  existing xterm CDN. Both are worth folding into a vendoring decision later.

## Testing

1594 passed, 63 skipped. ruff and mypy clean. Verified in a browser at 1280px
and 390px across dashboard, app detail, settings, system, diagnostics, add app,
terminal, docs, login, setup and `/design`.


compute_space/src/compute_space/web/helpers/static.py
```diff
@@ -0,0 +1,24 @@
...
+    Browsers aggressively cache static JS/CSS, so a deploy that ships a new
+    template + JS would otherwise leave returning visitors running stale JS
+    against new HTML.  Appending the file's mtime forces a fresh fetch.
+    """
+
+    def static_url(filename: str) -> str:
+        base = f"/static/{filename}"
+        try:
+            mtime = int((static_dir / filename).stat().st_mtime)
+        except OSError:
```
zplizzi: when would this happen? i don't like a silent degredation

compute_space/src/compute_space/web/routes/pages/design.py
```diff
@@ -0,0 +1,33 @@
+import attr
```
zplizzi: let's drop this file

compute_space/src/compute_space/web/static/img/deco/clouds.svg
zplizzi, line 1: make these individual svgs so that we can animate them if desired

compute_space/src/compute_space/web/templates/design.html
```diff
@@ -0,0 +1,156 @@
+{% extends "layout.html" %}
```
zplizzi: let's drop this too

compute_space/src/compute_space/web/setup_app.py
```diff
@@ -195,8 +196,14 @@ def create_setup_app(config: Config) -> Litestar:
     )
     static_router = create_static_files_router(path="/static", directories=[static_dir])
 
+    def _install_template_globals(app: Litestar) -> None:
+        engine = app.template_engine
+        if isinstance(engine, JinjaTemplateEngine):
```
zplizzi: i think we should make this an assert?

compute_space/src/compute_space/web/static/css/components.css
```diff
@@ -0,0 +1,1029 @@
...
+.btn--accent:hover {
+  background: var(--color-accent-alt-hover);
+}
+
+.btn--danger,
+.btn-danger {
+  background: var(--color-danger-bg);
+  border-color: var(--color-danger-border);
+  color: var(--color-danger-text);
+}
```
andrewlaack, line 333: Danger buttons lose their danger color on hover. `.btn:hover` at line 180 outranks this block and there is no `.btn--danger:hover`, so Remove and Delete buttons go grey on hover while the primary and accent variants keep their color.

compute_space/src/compute_space/web/static/css/components.css
```diff
@@ -0,0 +1,1029 @@
...
+  flex-wrap: wrap;
+  gap: var(--space-2);
+  margin: var(--space-4) 0;
+}
+
+.notice {
+  padding: var(--space-3) var(--space-4);
+  background: #e6f2ff;
+  border: var(--border-w) solid #b8d8f5;
+  font-size: var(--text-xs);
```
andrewlaack: `.notice` is declared again at line 761 with the same shared properties plus `margin` and `border-left-width: 3px`. The later block wins on everything, so this one never applies.

compute_space/src/compute_space/web/templates/setup.html
```diff
@@ -1,93 +1,94 @@
...
-  </form>
-  <script src="/static/js/username-validation.js"></script>
-  <script>
-    // Live client-side username validation lives in the shared
-    // /static/js/username-validation.js (mirrors the server-side
-    // validate_owner_username rule). The field is optional here, so an
-    // empty value is acceptable and falls back to the default 'owner'.
-    var usernameError = window.OpenHostUsername.usernameError;
+  </label>
+  <p class="field-error" id="username-error" style="display:none;"></p>
```
andrewlaack: This element still hides via `style="display:none"` and the JS toggles `style.display`, while the rest of the UI moved to the `hidden` property. `base.css:19` now sets `[hidden] { display: none !important }`, so switching this to `hidden` later would leave the `style.display` toggles unable to show it again. Same pattern in `add-app.js:22` and `app-detail.js`.

compute_space/src/compute_space/web/helpers/static.py
```diff
@@ -0,0 +1,28 @@
...
+    def static_url(filename: str) -> str:
+        path = static_dir / filename
+        try:
+            mtime = int(path.stat().st_mtime)
+        except OSError as exc:
+            # Only reachable when a template names a file that isn't there (a
+            # typo, or an asset deleted without updating its references). Fail
+            # loudly: silently emitting an un-cache-busted URL to a 404 means
+            # the page renders subtly broken instead of telling anyone.
+            raise RuntimeError(f"static_url({filename!r}): no such static file at {path}") from exc
```
andrewlaack: Just noting, this might be what we want, but a missing asset now 500s every page that references it, rather than 404ing the one asset.

compute_space/src/compute_space/web/templates/system.html
```diff
@@ -21,21 +23,24 @@ <h3>Swap <span class="info-tip" tabindex="0" role="note" aria-label="What is swa
...
   <tbody id="ports-body"></tbody>
 </table>
+</div>
 
-<h2 class="section-title">Logs</h2>
-<pre class="log-output" id="cs-logs" style="max-height: 120em; white-space: pre-wrap; word-break: break-all;"></pre>
+{% call section_head("Logs") %}{% endcall %}
+<pre class="log-output log-output--tall" id="cs-logs"></pre>
 
+<script src="{{ static_url('js/dom.js') }}"></script>
```
andrewlaack, line 54: `dom.js` is loaded here but `system.js` never calls it, still building markup with `innerHTML` and its own `escHtml`. Either a leftover include, or this script was meant to be converted with the others and got missed.

andrewlaack (commented):
Just a few comments.

andrewlaack (approved):

PR #300, opened 2026-08-20
Read cloudinabottle.toml with openhost.toml fallback

andrewlaack:
Apps now declare their manifest as cloudinabottle.toml. The loader looks for cloudinabottle.toml first and silently falls back to the legacy openhost.toml (no warning), so existing app repos keep working unchanged.

What changed:
Add MANIFEST_FILENAMES and find_manifest_path() in core/manifest.py, and route parse_manifest plus the two app-dir discovery sites (apps.py, default_apps.py) through it.
Test harness: find_manifest_dir accepts both names via find_manifest_path; consumer_app writes the canonical name.
Rename the in-repo app manifests and test fixtures to cloudinabottle.toml.
Update docs (README, docs/src, claude.md, the openhost-context skill, harness readme), the dashboard add_app page, and a migration comment to cloudinabottle.toml.
Tests: canonical/fallback/precedence resolution, vendored install for both names, and find_manifest_dir walk-up coverage; flow tests use the canonical name while dedicated tests retain openhost.toml fallback coverage.

The manifest is read-only in production (never rewritten), so this is purely a read-path change. Remote default apps that still ship openhost.toml keep working via the fallback.


zplizzi:
i'd be inclined to just call this `cloudinabottle.toml`. this isn't typed often and file names of that length aren't too uncommon - eg we have `.pre-commit-config.yaml` in the same folder. and if this is sitting in third party repos, it'd be nice for it to be unambiguous what it's for.

andrewlaack:
`cloudinabottle.toml` now

zplizzi (approved):
lgtm, thanks!

PR #301, opened 2026-08-20
Rename the oh CLI to bottle

andrewlaack:
Renames the compute-space CLI to bottle. The executable command and console script are bottle; the Python distribution is named cloud-in-a-bottle-cli (unique/descriptive for a possible PyPI publish). The env var is BOTTLE_INSTANCE and the config directory is ~/.cloud_in_a_bottle_cli.

What changed:
main.py: cappa command name is now bottle; class Oh becomes Bottle.
config.py: CONFIG_DIR is ~/.cloud_in_a_bottle_cli, with a read-fallback to the legacy ~/.openhost config that migrates forward on the next save; OH_INSTANCE becomes BOTTLE_INSTANCE.
pyproject.toml: the console script is bottle (root and compute_space_cli); the compute_space_cli distribution name is cloud-in-a-bottle-cli.
Updates every command example in help text, docs, README, and the openhost-context skill; install/upgrade commands reference the distribution name cloud-in-a-bottle-cli.
Tests: BOTTLE_INSTANCE resolution, the legacy OH_INSTANCE now being ignored, and config-dir fallback/migration coverage including the pre-multi-instance flat (url/token) legacy format; e2e helper and diagnostics tests updated.

Backward compatibility: existing logins keep working because load() reads ~/.openhost when ~/.cloud_in_a_bottle_cli is absent, then migrates forward on the next save.

Scope: the separate openhost self-host/router CLI is intentionally unaffected. Server-side ~/.openhost data paths (router data dir, PID dir) are unrelated to the client CLI config and are left as-is.

### DIFF
--- a/compute_space_cli/pyproject.toml
+++ b/compute_space_cli/pyproject.toml
@@ -1,5 +1,5 @@
 [project]
-name = "cb"
+name = "bottle"
 version = "0.1.0"
 description = "Cloud in a Bottle CLI — manage apps on your compute space"
 requires-python = ">=3.12"
@@ -11,7 +11,7 @@ dependencies = [
 ]
 
 [project.scripts]
-cb = "compute_space_cli.main:main"
+bottle = "compute_space_cli.main:main"
 
 [build-system]
 requires = ["hatchling"]
--- a/compute_space_cli/pyproject.toml
+++ b/compute_space_cli/pyproject.toml
@@ -1,5 +1,5 @@
 [project]
-name = "oh"
+name = "cb"
 version = "0.1.0"
 description = "Cloud in a Bottle CLI — manage apps on your compute space"
 requires-python = ">=3.12"
@@ -11,7 +11,7 @@ dependencies = [
 ]
 
 [project.scripts]
-oh = "compute_space_cli.main:main"
+cb = "compute_space_cli.main:main"
 
 [build-system]
 requires = ["hatchling"]
--- a/compute_space_cli/pyproject.toml
+++ b/compute_space_cli/pyproject.toml
@@ -1,5 +1,5 @@
 [project]
-name = "bottle"
+name = "cloud-in-a-bottle-cli"
 version = "0.1.0"
 description = "Cloud in a Bottle CLI — manage apps on your compute space"
 requires-python = ">=3.12"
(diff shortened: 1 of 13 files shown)


zplizzi:
i weakly prefer `bottle` for this? in the interest of standardizing on a single shortname, and `bottle` doesn't overlap with any other common CLI.

andrewlaack:
went with `bottle`

compute_space_cli/src/compute_space_cli/config.py
```diff
@@ -7,9 +7,15 @@
 import attr
 import tomli_w
 
-CONFIG_DIR = Path.home() / ".openhost"
+CONFIG_DIR = Path.home() / ".bottle"
```
zplizzi: i think full `.cloudinabottle` is better here - so it's clear what this folder is if found on a system?

compute_space_cli/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [project]
-name = "oh"
+name = "bottle"
```
zplizzi: the python project is maybe better called like `cloud-in-a-bottle-cli`, so if we ever were to publish it to pypi it would have a unique/descriptive name. the actual executable command can/should still be `bottle`.

compute_space_cli/src/compute_space_cli/config.py
zplizzi: or actually if this is specifically for the cli, maybe `.cloud_in_a_bottle_cli/`?

zplizzi: idk up to you

zplizzi (approved):
lgtm with minor thoughts on secondary names

compute_space_cli/src/compute_space_cli/config.py
andrewlaack: Went with `~/.cloud_in_a_bottle_cli` — CLI-specific and clear on a system. Legacy `~/.openhost` is still read as a fallback and migrated forward on the next save.

compute_space_cli/pyproject.toml
andrewlaack: Renamed the distribution to `cloud-in-a-bottle-cli` (unique/descriptive for a possible PyPI publish); the executable command stays `bottle`. Updated the install/upgrade commands accordingly.

PR #302, opened 2026-08-20
Add wipe data & restart button

samuelren-imbue:
Wipes the non-repo data, then rebuilds
Also renamed some buttons to be shorter

Stop App -> Stop
Reload App -> Reload
Edit App -> Edit


compute_space/src/compute_space/core/apps.py
```diff
@@ -1124,6 +1126,47 @@ def reload_app_background(app_id: str, repo_path: str, config: Config) -> None:
         db.close()
 
 
+def wipe_data_restart_background(app_id: str, config: Config) -> None:
+    """Stop an app, remove its data, and rebuild it from its checked-out source."""
+    db = sqlite3.connect(config.db_path, check_same_thread=False)
+    db.row_factory = sqlite3.Row
+    db.execute("PRAGMA journal_mode=WAL")
+    try:
+        app_row = db.execute("SELECT * FROM apps WHERE app_id = ?", (app_id,)).fetchone()
```
zplizzi: nit: can you make a `find_app_by_id` similar to `find_app_by_name`.

compute_space/src/compute_space/web/routes/api/apps.py
```diff
@@ -855,6 +856,56 @@ async def reload_app_after_oauth(
     )
 
 
+@post(
+    "/wipe_data_restart/{app_id:str}",
+    status_code=202,
+    guards=[require_owner_auth],
+    raises=[ValidationException, NotFoundException, ConflictException, ServiceUnavailableException],
+)
+async def wipe_data_restart(
```
zplizzi, line 683: i think most of this could go into a `core` function - generally it's better to keep logic out of the routes as much as possible, to make it easier to test and reuse.

zplizzi (approved):

compute_space/src/compute_space/web/routes/api/apps.py
samuelren-imbue: a lot of the other routes use internal functions like _reload_app_impl, should these get moved to core logic as well?

samuelren-imbue: @zplizzi

zplizzi: yeah likely, if the web-related stuff can be cleanly separated out. core shouldn't have anything litestar or web related (eg no Response objects, no litestar exceptions, etc). doesn't necessarily need to happen here tho unless helpful

compute_space/src/compute_space/web/routes/api/apps.py
```diff
@@ -524,11 +462,14 @@ async def app_diagnostics(
     ``?download=1`` adds a Content-Disposition header so browsers save the JSON
     to a timestamped file instead of rendering it inline.
     """
-    app_row = _resolve_app(app_id, db)
+    try:
+        app_row = resolve_app(app_id, db)
+    except AppError as exc:
+        raise _to_http_error(exc) from exc
```
samuelren-imbue, line 468: Moving functions into core has left a lot of this repeated. Should we have a wrapper for this? It would be kind of confusing indirection. @zplizzi

zplizzi: i'd probs move this into a custom exception handler on the Router at the bottom of apps.py. eg something like:
```

def handle_app_error(request, exc: AppError) -> Response:
    # (adapt this to your specific exception subtypes and return codes etc)
    return Response(
        content={"error": str(exc)},
        status_code=HTTP_500_INTERNAL_SERVER_ERROR,
    )


my_router = Router(
    path="/items",
    route_handlers=[...],
    exception_handlers={AppError: handle_app_error},
)
```

then you can just annotate each route with the specific AppError subclasses that they raise (which is better than just annotating with the generic AppError) and i think everything will work out.

zplizzi:
otherwise this looks like a good cleanup, thanks!

compute_space/src/compute_space/web/routes/api/apps.py
samuelren-imbue: does annotating with the AppError subclasses mean putting it into raises=[]? can litestar handle that and turn it into openapi specs?

zplizzi: hmm i was thinking so but maybe not, can you try?

samuelren-imbue: turns out no, litestar can only directly take http exceptions

zplizzi: either of these approaches seems ok: https://claude.ai/share/c50959cc-245d-457f-aab4-b5ef5598724e

samuelren-imbue: this works for the end result, but i think anything here is kind of unclear as to where errors come from and how to track them in the code itself. maybe it's not that big an issue though?

zplizzi: i think just pick something that seems good to you and let's get this merged :)

PR #303, opened 2026-08-20
Strip Authorization before forwarding to apps, only when it is an OpenHost credential 

andrewlaack:
The router forwarded the inbound Authorization header verbatim to apps. An owner API token grants full API access, so any app opened with one could capture and replay it (reported in Slack).

Fix: strip the Authorization header before forwarding only when its bearer is a credential the router itself consumed, i.e. a valid OpenHost API token or app token. An app's own bearer (e.g. a JWT an app issues to its SPA and sends back through the router, like lyftr) is not ours and is forwarded untouched, so apps that do their own bearer auth keep working. Owner API tokens and app tokens are never leaked to backends, on both the app-subdomain proxy and the cross-app service proxy.

A new authorization_is_openhost_credential() check (valid api/app token) drives a strip_authorization flag threaded through proxy_http_request / proxy_websocket_request. Apps still get caller identity via X-OpenHost-Is-Owner and X-OpenHost-Consumer-*.

Tested on a live instance: owner API token no longer reaches the app while X-OpenHost-Is-Owner still does; lyftr (localStorage-JWT app) works again; cross-app service calls still work; WS handshake works with a bearer and unauthenticated WS is still refused. Browser-checked a spread of catalog apps (redlib, navidrome, calibre-web, openweb-ui, pihole, miniflux, uptime-kuma, forgejo, vaultwarden, md-notes, lyftr).

Tests: sanitizer unit tests (strip when flagged, keep app-owned bearer otherwise), an integration test that an owner token never reaches the app, one that an app-owned bearer is forwarded, and a real-backend service-proxy test that the consumer app token is stripped while X-OpenHost-Consumer-Id arrives.


compute_space/src/compute_space/web/helpers/proxy.py
```diff
@@ -288,10 +302,13 @@ async def proxy_websocket_request(
     target_port: int,
     extra_headers: Iterable[tuple[str, str]] = (),
     override_path: str | None = None,
+    strip_authorization: bool = False,
```
zplizzi, line 305: would true be the better/safer default?

zplizzi (approved):
this seems fine, altho i just realized an alternative that might be simpler+more robust would be to switch from using `Authorization` to `X-Openhost-Authorization` or something like that, and then always stripping those along with all the other X-Openhost headers. then there's no chance of collision with apps using an `Authorization` header.

andrewlaack:
> switch from using Authorization to X-Openhost-Authorization or something like that

I'd prefer this. Making this a draft until then.

PR #306, opened 2026-08-20
Bound build swap to the memory limit instead of leaving it unlimited

kilo-gi:
## What

`build_image` passed `--memory-swap=-1` alongside `--memory`, giving each app build **unlimited** host swap. This sets `--memory-swap` equal to `--memory` so builds get **zero swap**.

`--memory-swap` is the combined memory+swap ceiling ([podman-build.1](https://docs.podman.io/en/latest/markdown/podman-build.1.html)). Setting it equal to `--memory` makes that ceiling the memory limit itself, so there is no swap headroom — a build stays within `build_memory_mb` or OOMs. (Left unset, podman defaults it to `2 × memory`, which would allow `memory_mb` of swap.)

## Why

Follows the pass through the catalog setting per-app `build_memory_mb`. Per review (Zack), builds should not touch swap at all — a hard memory ceiling with no disk spill.

## Caveat

A build needing more than `build_memory_mb` now OOMs immediately (no swap spill). If any catalog app's `build_memory_mb` is set too tight for its actual build footprint, its build will fail — worth watching the first heavy build after this lands.

## Tests

Updated `test_build_image_applies_memory_limit` to assert the argv contains `--memory=512m` and `--memory-swap=512m`. `test_build_image_omits_memory_flag_when_unset` still holds. All `build_image` tests pass.

### DIFF
--- a/compute_space/src/compute_space/core/containers.py
+++ b/compute_space/src/compute_space/core/containers.py
@@ -192,13 +192,13 @@ def build_image(
         dockerfile_path,
     ]
     if memory_mb is not None:
-        # Cap build RAM at the app's declared limit for isolation. We pass only
-        # --memory and leave --memory-swap unset, so podman applies its default
-        # of memory-swap == 2*memory: swap is bounded to the same size as the
-        # memory limit rather than being unlimited. A build that needs more
-        # than memory_mb can spill up to memory_mb into swap before OOMing,
-        # without letting a "small" app hog unbounded host swap.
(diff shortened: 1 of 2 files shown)


zplizzi:
i thought we agreed to just disable swap entirely for builds? i'm worried this would lead to slow builds for non-obvious reasons.

kilo-gi:
You want to set swap limit to the same as memory limit?

Sure, that's straightforward.

PR #308, opened 2026-08-21
Truncate build error

samuelren-imbue:
In the description table of the detail page, only show the final line of the error (will be docker build step typically) capped at 150 char. 
Also, added a websocket to add the "error" row when an error happens, previously only updated on page refresh.


zplizzi (changes requested):
i think we should probably just record/show "build failed" or something like that and not try to put the log line in here? in the logs panel the full logs are available. also i'd be inclined to just poll every ~1s instead of a websocket, just much simpler code wise. we only do websocket for the logs to avoid reloading the full log output each poll.

samuelren-imbue:
> i think we should probably just record/show "build failed" or something like that and not try to put the log line in here? in the logs panel the full logs are available. also i'd be inclined to just poll every ~1s instead of a websocket, just much simpler code wise. we only do websocket for the logs to avoid reloading the full log output each poll.

I would like the last line to show, since it's which Docker step failed, rather than like some line of an exception trace. I think it could be immediately useful. Will remove websocket.

zplizzi (approved):

PR #309, opened 2026-08-21
Local dev stack: justfile, random port, and macOS archive-tier fix

zplizzi:
- don't dump a juicefs failure when running locally on mac (bc the binaries are linux only). instead just raise a warning and set the archive backend state to an error message (is this right?). we could get mac binaries but seems like we'd want to handle unsupported platforms cleanly regardless so this seems good.
- allow running the compute space locally with justfile and by default run on a random port to avoid collisions.

### DIFF
--- a/compute_space/src/compute_space/core/archive_backend.py
+++ b/compute_space/src/compute_space/core/archive_backend.py
@@ -49,6 +49,7 @@ from compute_space.core.domains import primary_domain_or_none
 from compute_space.core.logging import logger
 from compute_space.core.pinned_binary import get_pinned_binary
 from compute_space.core.pinned_binary import install_pinned_binary
+from compute_space.core.pinned_binary import is_linux_host
 
 # Name of the systemd unit that manages the JuiceFS FUSE mount.
 # Installed by ansible (disabled); enabled by compute_space when the
@@ -290,7 +291,7 @@ def format_local_volume(config: Config, juicefs_volume_name: str) -> None:
         _format_meta_dsn(config),
         juicefs_volume_name,
     ]
-    logger.info("Running juicefs format (local file backend) at %s", store_dir)
+    logger.info("Running juicefs format (local file backend) at {}", store_dir)
     result = subprocess.run(cmd, capture_output=True, text=True, timeout=60)
     if result.returncode != 0:
         raise RuntimeError(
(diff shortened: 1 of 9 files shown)


compute_space/src/compute_space/core/pinned_binary.py
```diff
@@ -102,12 +110,18 @@ def get_pinned_binary(name: str) -> PinnedBinary:
 
 def install_pinned_binary(binary: PinnedBinary, dest_path: str) -> None:
     """Download + verify sha256 + extract ``binary`` to ``dest_path``.  Idempotent."""
+    if not is_linux_host():
+        # Checked before the "already installed" shortcut: a Linux binary left on disk by an
+        # earlier run looks installed but dies with "Exec format error" when run.
+        raise UnsupportedHostPlatformError(
+            f"{binary.name} is only pinned as a Linux release; this host is {os.uname().sysname}."
+        )
```
andrewlaack, line 118: `_ensure_coredns_binary` at `web/start.py:132` calls this without an `is_linux_host()` guard, so on a non-Linux host with `coredns_enabled=True` and no `coredns` on PATH, boot crashes with this exception unhandled. That is the same failure mode the PR fixes on the JuiceFS side.

compute_space/src/compute_space/core/archive_backend.py
```diff
@@ -655,6 +656,16 @@ def attach_on_startup(config: Config, db: sqlite3.Connection) -> None:
...
     state = read_state(db)
+    if state.backend in ("local", "s3") and not is_linux_host():
+        # JuiceFS is pinned as a Linux release only, so the archive tier can't run here (local
+        # dev on macOS).  Everything else works; say so once instead of raising on every boot.
+        logger.warning(
+            "Archive tier disabled: JuiceFS has no {} build, so the '{}' backend can't be brought up.",
+            os.uname().sysname,
+            state.backend,
+        )
+        _set_state_message(db, f"Archive tier unavailable: no JuiceFS build for {os.uname().sysname}.")
```
andrewlaack: Nit: log says "disabled", state message says "unavailable". Also `os.uname().sysname` is called twice.

justfile
```diff
@@ -0,0 +1,10 @@
+_default:
+    @just --list
+
+# run a local compute space on a random port (extra args pass through, eg --port 8080 --default-apps)
```
andrewlaack: Nit: `--port 8080` is the default this PR is moving away from. Perhaps pick a neutral example like `--port 9000`.

justfile
```diff
@@ -0,0 +1,10 @@
+_default:
```
andrewlaack, line 4: `just` isn't in `pyproject.toml` or `pixi.lock`, so first-time contributors hit `command not found`.

andrewlaack (approved):

justfile
zplizzi: added to dev group and a comment in the justfile

compute_space/src/compute_space/core/pinned_binary.py
zplizzi: i think crashing is appropriate there - if coredns can't start, we probs shouldn't start the space at all. vs the archive backend not starting is less critical - it just gets reported as unhealthy which is already a handled case.

PR #310, opened 2026-08-21
Add clean build cache button

samuelren-imbue:
Build cache now shows size of cache excluding active images
Clean cache button with confirmation screen


compute_space/src/compute_space/core/containers.py
```diff
@@ -502,7 +503,7 @@ def container_image_storage_bytes() -> int | None:
             return None
         for row in json.loads(result.stdout):
             if row.get("Type") == "Images":
-                return int(row["RawSize"])
+                return int(row["RawSize"]), int(row["RawReclaimable"])
```
andrewlaack: Hard key access inside the broad `except Exception`. If a podman version omits `RawReclaimable`, the function returns None and `build_cache_bytes` disappears too. This would be a regression from the pre-PR behavior. Suggest `int(row.get("RawReclaimable", 0))`.

compute_space/src/compute_space/web/static/js/system.js
```diff
@@ -79,6 +79,49 @@ function renderListeningPorts(data) {
 
 // ─── Storage Status ───
 
+// Message state lives outside the DOM because renderStorageStatus rebuilds the
+// table (and thus the message span) whenever storage data is re-fetched.
+var cleanCacheMsg = {cls: 'muted', text: ''};
+var cleanCacheInFlight = false;
+var cleanCacheFreeBytes = null;
```
andrewlaack: Nit: `cleanCacheFreeBytes` holds a prunable estimate, not bytes freed. `cleanCacheReclaimableBytes` would match the API field.

andrewlaack (approved):

compute_space/src/compute_space/core/containers.py
samuelren-imbue: making it return none instead of 0, makes more sense to me

PR #312, opened 2026-08-24
Use provider-neutral deploy copy

andrewlaack:
Changes the dashboard deploy card to refer to the user's preferred Git hosting provider instead of GitHub and Claude.


zplizzi:
how about "From your own repo."? And the docs and setup one could be "Learn about your space and creating new apps." or something.

PR #315, opened 2026-08-25
Dual-name env vars: OPENHOST_ (legacy) + BOTTLE_ (new)

kilo-gi:
## What

The project was renamed **OpenHost → Cloud in a Bottle**, but the env-var contract still used `OPENHOST_*` names. This keeps every legacy name for backward compatibility and adds a parallel `BOTTLE_*` set. `BOTTLE_` is canonical going forward and wins when both are set. Writers that cross a version boundary (self-update, agent forwards) set **both** names so either side resolves the value.

## Changes

**App + daemon contract**
- App container env: `run_container` mirrors every var to a `BOTTLE_` twin via `add_bottle_env_aliases()` (`core/env_aliases.py`), after host→container path translation.
- Daemon config: `load_config()` appends an `EnvLoader(prefix="BOTTLE_")` so `BOTTLE_<FIELD>` overrides `OPENHOST_<FIELD>`. Config-file path resolution shared via `router_config_path_from_env()` (prefers `BOTTLE_ROUTER_CONFIG`), used by both `load_config` and `first_boot`.
- CLIs: `up.py` sets both `*_ROUTER_CONFIG`; `compute_space_cli` prefers `BOTTLE_INSTANCE` over `OH_INSTANCE`.

**Deploy + system agent** (found by re-grepping every `OPENHOST_` occurrence)
- systemd unit: `build_openhost_service_unit` + `openhost.service.j2` set `Environment=BOTTLE_ROUTER_CONFIG` too. **Migration `v0011`** rewrites the installed unit + `daemon-reload` so existing hosts pick it up on self-update (mirrors `v0008`).
- Data-dir override: `paths.py` reads `BOTTLE_DATA_DIR` (preferred) / `OPENHOST_DATA_DIR`; every forwarder (`start.py`, system-agent client, `detach`, updater `launcher`) exports both.
- Detached-apply marker: `detach.py` sets/reads `BOTTLE_APPLY_DETACHED` + legacy.

## Verification

Every `OPENHOST_` occurrence in the tree was grep-classified; the only unpaired residual is intentional (prose/docs, test-deploy plumbing, bundled demo apps on the kept legacy names, image tags / unit *names* / path constants, and `apps.py` internal reads of `provision_data`'s own dict). A runtime proof confirms all container-env vars have equal-valued `BOTTLE_` twins and `BOTTLE_` config overrides win.

Tests: alias helper, container argv twins, `load_config` precedence, `BOTTLE_INSTANCE`, migration `v0011`, data-dir env precedence, `first_boot` BOTTLE_ path, detach both-names. Full suites green (`compute_space` + `compute_space_cli` + `routerd_cli`: 1517 passed; `openhost_system_agent` green); project-wide mypy + ruff clean.

## Still on the (kept) legacy names — intentional
Human-written docs (`docs/src/*.md`), e2e shell plumbing, and bundled demo apps under `apps/` — all work via the compatibility names.

### DIFF
--- a/ansible/files/openhost.service.d/20-bottle-router-config.conf
+++ b/ansible/files/openhost.service.d/20-bottle-router-config.conf
@@ -1,0 +1,12 @@
+# Managed by OpenHost; do not edit by hand. Kept byte-identical between
+# ansible/files/openhost.service.d/20-bottle-router-config.conf and the
+# BOTTLE_ROUTER_CONFIG_DROPIN constant in openhost_system_agent's v0011 migration
+# (a test enforces this).
+#
+# OpenHost -> Cloud in a Bottle rename: advertise the router config-file path under
+# the new BOTTLE_ROUTER_CONFIG name. The main unit already sets the legacy
+# OPENHOST_ROUTER_CONFIG and the router reads either (BOTTLE_ preferred), so this is
+# additive — Environment= directives accumulate across drop-ins, layering the new
+# name on without touching the baseline unit.
+[Service]
+Environment=BOTTLE_ROUTER_CONFIG=/home/host/.openhost/local_compute_space/config.toml
--- a/ansible/tasks/install_openhost_units.yml
+++ b/ansible/tasks/install_openhost_units.yml
@@ -83,6 +83,15 @@
     mode: "0644"
   register: journal_dropin_file
 
+- name: Install openhost bottle-router-config drop-in
+  become: yes
+  become_user: root
+  copy:
+    src: files/openhost.service.d/20-bottle-router-config.conf
+    dest: /etc/systemd/system/openhost.service.d/20-bottle-router-config.conf
+    mode: "0644"
+  register: bottle_router_config_dropin_file
+
 # JuiceFS archive-tier mount: a dedicated systemd service with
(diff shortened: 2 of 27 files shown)


kilo-gi:
I'm not 100% sure what should be done about the migration, but for now I'll leave as-is and the rest should be reviewable.

Posted in slack about the migration and we'll discuss there.

zplizzi (commented):
can you pull out just the stuff for the env vars on the openhost->app boundary into a separate PR (and remove the stuff to deal w import cycles)? ie the ones that get set in app containers. the rest of it feels pretty complicated and not super urgent - users will never see or interact with those.

kilo-gi:
ok, splitting into https://github.com/cloud-in-a-bottle/cloud-in-a-bottle/pull/322 and https://github.com/cloud-in-a-bottle/cloud-in-a-bottle/pull/323

I'm not entirely sure what you mean by 'import cycles' - I moved the migration and using BOTTLE_ROUTER_CONFIG internally into the second PR. I don't think that's super complex - Like yes it's a bigger diff, but it's ~all functionally just moving to handle two values on the path that was previously only handling one.

PR #321, opened 2026-08-27
move coredns into a router-provided service

zplizzi:
add a builtin DNS service provider, provided on the service interface, using coredns. reworks the plumbing to access this via the service interface vs directly.


zplizzi:
@andrewlaack this one isn't done yet but is an example of how the previous steps will be used.

zplizzi:
Split the parts of this that stand on their own into #335 — the `InternalDnsProvider` collapse and the cross-import removal, with records held in memory and re-published each boot instead of going through a service.

Leaving this one open in case we want to come back to the service interface.

PR #322, opened 2026-08-27
Dual-name env vars passed into apps: OPENHOST_ + BOTTLE_

kilo-gi:
Part 1 of splitting #315.

The project was renamed OpenHost -> Cloud in a Bottle. This is the **app-facing** slice: every env var stamped into an app container is now exposed under both the legacy `OPENHOST_` name and the new `BOTTLE_` name, so deployed apps can read either.

## What changed
- New dependency-free `core.env_aliases` module with `add_bottle_env_aliases`, which mirrors each `OPENHOST_*` var to a `BOTTLE_*` twin without clobbering an explicit `BOTTLE_*` value. Re-exported from `core.data` (the historical home of the mirror boundary).
- `run_container` applies it after host->container path translation and after `LOCAL_PORT` is set, so the twins carry the final in-container values (path vars get the translated container path, not the host path).

## Scope
Deliberately small and mergeable on its own. The internal daemon/agent config renames, the data-dir/updater env handling, `BOTTLE_ROUTER_CONFIG`, and the systemd-drop-in migration all live in the stacked follow-up PR.

Replaces #315 (to be closed).

### DIFF
--- a/compute_space/src/compute_space/core/containers.py
+++ b/compute_space/src/compute_space/core/containers.py
@@ -36,6 +36,30 @@ from compute_space.core.manifest import PortMapping
 
 CONTAINER_ROOT = "/data"
 
+# Env-var naming contract.  The project was renamed OpenHost -> Cloud in a
+# Bottle; every OPENHOST_* variable stamped into an app container is now also
+# exposed under BOTTLE_*.  The legacy names are kept indefinitely for backward
+# compatibility.
+LEGACY_ENV_PREFIX = "OPENHOST_"
+ENV_PREFIX = "BOTTLE_"
+
+
+def add_bottle_env_aliases(env: dict[str, str]) -> dict[str, str]:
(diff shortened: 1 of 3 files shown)


compute_space/src/compute_space/core/data.py
```diff
@@ -8,6 +8,15 @@
 
 from compute_space.core.containers import ROUTER_GATEWAY_HOST
 from compute_space.core.containers import ROUTER_LOOPBACK_HOST
+
```
zplizzi: this doesn't really make sense to me, is it necessary?

compute_space/src/compute_space/core/env_aliases.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
```
zplizzi: i don't understand the import cycle. can we just put this in containers? or otherwise do this more elegantly than making a new file for this

kilo-gi: oh yeah, we can just put it all in containers. Done.

PR #326, opened 2026-08-27
use asyncio more consistently in core

zplizzi:
use async more consistently in compute_space/core. makes things easier to reason about than having a mix of async and sync imo.

### DIFF
--- a/compute_space/src/compute_space/core/tls/cert_api_client.py
+++ b/compute_space/src/compute_space/core/tls/cert_api_client.py
@@ -107,46 +107,46 @@ class CertApiClient:
     """Thin synchronous client over the broker's REST API.
 
     Construct via ``CertApiClient.create(base_url, token_provider)`` in production;
-    tests inject an ``httpx.Client`` backed by a MockTransport plus a StaticTokenProvider.
+    tests inject an ``httpx.AsyncClient`` backed by a MockTransport plus a StaticTokenProvider.
     """
 
-    http_client: httpx.Client
+    http_client: httpx.AsyncClient
     token_provider: TokenProvider
 
     @classmethod
     def create(cls, base_url: str, token_provider: TokenProvider, timeout: float = 30.0) -> CertApiClient:
-        http_client = httpx.Client(
+        http_client = httpx.AsyncClient(
             base_url=base_url.rstrip("/"),
             timeout=timeout,
         )
         return cls(http_client=http_client, token_provider=token_provider)
 
-    def __enter__(self) -> CertApiClient:
+    async def __aenter__(self) -> CertApiClient:
         return self
 
-    def __exit__(
+    async def __aexit__(
         self,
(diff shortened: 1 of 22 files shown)


compute_space/src/compute_space/web/routes/api/domains.py
```diff
@@ -234,9 +223,8 @@ async def remove_domain(
     if not remove_record(db, name):
         raise NotFoundException(detail="domain not found")
     if removed is not None and not removed.mdns:
-        # Drop the zone from CoreDNS so it stops answering for the removed public domain.  Off the
-        # event loop — the restart blocks on a terminate+wait(3s).
-        await anyio.to_thread.run_sync(reload_coredns_for_domains, config, db)
+        # Drop the zone from CoreDNS so it stops answering for the removed public domain.
+        await reload_coredns_for_domains(config, db)
```
zplizzi, line 227: this is a behavior change but seems fine, this should be quickish

zplizzi:
@andrewlaack if you only have small feedback feel free to just make changes and merge

compute_space/src/compute_space/core/tls/util.py
```diff
@@ -60,32 +60,41 @@ def _wait_for_txt_propagation(
...
+            f"@{resolver}",
+            qname,
+            "TXT",
+            "+short",
+            "+timeout=5",
+            "+tries=1",
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.DEVNULL,
+        )
+        stdout, _ = await asyncio.wait_for(proc.communicate(), timeout=10)
```
andrewlaack: On timeout, `dig` is neither killed nor reaped. Could we kill and await it before returning?

compute_space/src/compute_space/core/tls/util.py
```diff
@@ -172,25 +181,27 @@ def _acquire_cert_dns01(
                 )
 
                 # Wait for CoreDNS to pick up the zone file change (reload interval = 2s)
-                time.sleep(3)
+                await asyncio.sleep(3)
```
andrewlaack: Cancellation after writing TXT records skips cleanup. Could cleanup move into a `finally` block?

compute_space/src/compute_space/core/tls/cert_api_client.py
```diff
@@ -107,46 +107,46 @@ class CertApiClient:
...
     """
 
-    http_client: httpx.Client
+    http_client: httpx.AsyncClient
     token_provider: TokenProvider
 
     @classmethod
     def create(cls, base_url: str, token_provider: TokenProvider, timeout: float = 30.0) -> CertApiClient:
-        http_client = httpx.Client(
+        http_client = httpx.AsyncClient(
```
andrewlaack, line 118: This client is now async, but the docstring still describes it as synchronous.

PR #327, opened 2026-08-27
cleanup service interface code, add builtins served by the router, and add a way to call services from within python

zplizzi:


zplizzi:
@andrewlaack i spent quite a while on this but some of the logic / types still feel kinda messy/duplicative/inelegant. if you get a chance to look through it and see anything that could be improved, i'd appreciate it. otherwise feel free to just review.

zplizzi:
also if it just doesn't make sense lmk and i can keep working on it. the main goal was to make a generic interface for having services served by the router, and also to be able to call them directly from within the router without doing a loopback HTTP request (which wouldn't work without caddy anyways and so caused boot ordering issues with DNS).

zplizzi:
@andrewlaack i'd still like to review/merge this one

compute_space/src/compute_space/core/service_interface/services.py
```diff
@@ -0,0 +1,133 @@
...
+        for svc in manifest.provides_services_v2:
+            db.execute(
+                "INSERT OR REPLACE INTO service_providers_v2 (service_url, app_id, service_version, endpoint) VALUES (?, ?, ?, ?)",
+                (svc.service, app_id, svc.version, svc.endpoint),
+            )
+            existing_default = db.execute(
+                "SELECT 1 FROM service_defaults WHERE service_url = ?",
+                (svc.service,),
+            ).fetchone()
+            if not existing_default:
```
andrewlaack: Could we preserve an explicit router default across app re-registration? Since selecting the router removes the row, startup or reload makes the app default again.

compute_space/src/compute_space/web/routes/api/services_v2.py
```diff
@@ -62,99 +34,36 @@ class RemoveDefaultRequest:
...
-    db.commit()
+    """Remove the default provider for a service, handing it back to the builtin if there is one."""
+    clear_default(data.service_url, db)
     return OkResponse(ok=True)
 
 
 api_services_v2_routes = Router(
     path="/",
-    route_handlers=[list_services_v2, discover_providers, list_defaults, set_default, remove_default],
+    route_handlers=[list_services_v2, discover_providers, set_default_route, remove_default],
```
andrewlaack, line 68: `cross_app_services.md` should to be updated here as it still documents the removed GET endpoint and old clearing behavior.

compute_space/src/compute_space/core/oauth.py
```diff
@@ -43,17 +44,18 @@ async def get_oauth_token(
...
 
     # Forge an app-scoped grant. We bypass the v2 service proxy (which would normally do the per-provider
-    # filter + provider_app_id strip) by going straight to the loopback port, so we hand-build the same
+    # filter + provider_app_id strip) by going straight to the provider, so we hand-build the same
     # post-filter shape the proxy would produce. The oauth service only honours app-scoped grants
     # (see services/oauth/openapi.yaml).
     grant_payload = {"provider": provider, "scopes": list(scopes)}
     permissions_header = json.dumps([{"grant": grant_payload, "scope": "app"}])
-    url = f"http://127.0.0.1:{port}{endpoint.rstrip('/')}/token"
+    client, base_url = client_for(oauth_provider.target, 5)
```
andrewlaack, line 51: This should probably use `ROUTER_APP_ID` and `ROUTER_APP_NAME` too because OAuth still sends the legacy router identity.

openhost_system_agent/README.md
```diff
@@ -65,7 +65,7 @@ that ref.
            run("systemctl", "restart", "some-service")
    ```
 
-3. Register the migration in `registry.py` by appending an instance to `REGISTRY`.
+3. Register the migration in `services.py` by appending an instance to `REGISTRY`.
```
andrewlaack: Both references should remain `registry.py` bc the migrations package has no `services.py`.

andrewlaack (approved):

compute_space/src/compute_space/core/service_interface/services.py
zplizzi: you can't easily do it explicitly bc defaults has a foreign key to the app table and the router isn't in the app table. but why would startup/reload make the app default again? i think we only write the defaults table if the user explicitly picks an alternative default in the settings page.

zplizzi: er nvm we do write the default on eg app install. i'm gonna change this to never writing the defaults table unless the user explicitly picks something, and having a good implicit default ordering that is used otherwise.

zplizzi: (done)

PR #328, opened 2026-08-28
Remove the installer v2 service

zplizzi:
we didn't end up using this, and this will get incorporated into a broader "openhost admin interface" service.


compute_space/src/compute_space/db/versioned/migrations/v0014_drop_apps_installed_by.py
```diff
@@ -0,0 +1,10 @@
+"""v14: drop ``apps.installed_by``.  Body in v0014_drop_apps_installed_by.sql."""
```
andrewlaack: nit: Should probably avoid a file-level docstring in this case ([style guide](https://github.com/cloud-in-a-bottle/cloud-in-a-bottle/blob/135c3b070ac8882d650b07942b284d1490b2be95/style_guide.md#L5)).

andrewlaack (approved):

compute_space/src/compute_space/db/versioned/migrations/v0014_drop_apps_installed_by.py
zplizzi: Good catch — dropped it. The rationale lives in the .sql body already.

zplizzi: ah god i don't like claude impersonating me

PR #333, opened 2026-08-31
image: distributable VM image + release workflow for easy self-host installs (OH-275)

kilo-gi:
Linear: [OH-275](https://linear.app/imbue/issue/OH-275/build-vm-image-that-we-can-distribute-for-easy-self-host-installs)

Combines the VM-image build and the GitHub release workflow into one PR.

## `image/` — the image builder

`image/build.sh` boots the Ubuntu 24.04 cloud image once under KVM with a cloud-init seed that runs the existing, tested `scripts/provision.sh`, then freezes the disk into a **qcow2** (QEMU) and **OVA** (VirtualBox), x86_64. No Packer — it exercises the real deploy path, not an offline-chroot approximation.

**Zero-config first boot:** HTTP-only, bound `0.0.0.0` — no TLS/CoreDNS/Caddy, **no domain required**. Boot → dashboard at `http://<vm-ip>:8080`.

**Open claiming by default (no token).** A fixed token baked into every image is a public non-secret — friction without protection. Since the image is private behind NAT (nothing reaches `:8080` without the user's own port-forward), the image sets `claim_token_required=false`. `claim_token_required` is now a config var (default `true`, so **no change for any existing deploy**); pass `--claim-token` to bake one for a customized image.

**No SSH key baked in.** Access is console-only until the owner adds their own key (`--ssh-pubkey` is opt-in for custom builds).

**Grow-to-fill.** `image/seal.sh` installs `openhost-prepare.service`, a systemd oneshot that grows the root filesystem to fill whatever disk the image is installed onto (device auto-detected → works on QEMU `vda` and VirtualBox `sda`). "Install size" = the disk you give the VM; default floor 20G. Done with systemd, not cloud-init, because a datasource-less appliance never runs cloud-init. It also generalizes the image: fresh machine-id and regenerated SSH host keys per install.

`scripts/provision.sh` gains `--bind-host`, `--claim-token`, `--swap-size`, and `--open-claim` passthroughs so the build reuses the single tested provisioning path.

## `.github/workflows/image-release.yml` — the release flow

Manually-triggered (`workflow_dispatch`) job: takes a git tag, runs `image/build.sh` for it, and publishes the qcow2 + OVA as a GitHub release (draft by default). Shells straight out to the build script — no separate recipe.

> Note: GitHub-hosted `ubuntu-latest` may not expose `/dev/kvm`; the workflow warns and falls back to TCG, which is much slower. Worth confirming KVM availability (or a larger/self-hosted runner) on the first real run.

## Verification (on a KVM host, booting the produced image fresh)

| Check | 60G disk (resized) | 20G native floor |
|---|---|---|
| Root FS grew | 58G ✅ | 19G (no-op) ✅ |
| `openhost-prepare` | success ✅ | success ✅ |
| SSH host keys regenerated | ✅ | ✅ |
| Fresh machine-id | ✅ | ✅ |
| Dashboard health | 200 ✅ | 200 ✅ |

Open-claim verified through the real ansible render: unset → `claim_token_required = true`, `-e claim_token_required=false` → `false`.

## Not included (intentional)

- No binary artifacts committed — the release workflow publishes versioned, downloadable images.
- macOS HVF build path and the app-subdomain DNS/UX polish are follow-ups.

### DIFF
--- a/.github/workflows/image-release.yml
+++ b/.github/workflows/image-release.yml
@@ -84,8 +84,9 @@ jobs:
           - \`*.qcow2\` — QEMU/KVM/libvirt (and most cloud importers)
           - \`*.ova\` — VirtualBox / VMware
 
-          Boot it, then reach the dashboard at \`http://<vm-ip>:8080\`.
-          See \`image/build.sh\` and \`docs/\` for defaults (claim token, console login)."
+          Boot it, then reach the dashboard at \`http://<vm-ip>:8080\` and claim it
+          (claiming is open by default — the image is private behind NAT).
+          See \`image/build.sh\` for defaults (console login, disk sizing)."
           )
           if [ "${{ inputs.draft }}" = "true" ]; then
             create_args+=(--draft)
--- a/.github/workflows/image-release.yml
+++ b/.github/workflows/image-release.yml
@@ -1,0 +1,94 @@
+name: image-release
+
+# Build the OpenHost VM images (qcow2 + optional VirtualBox OVA) for a chosen
+# git tag and publish them as a GitHub release. Manually triggered for now —
+# run it from the Actions tab once the version tag has been pushed.
+#
+# The build recipe (image/build.sh) boots the tag's provision.sh once under
+# QEMU/KVM and freezes the result; the app code baked into the image is cloned
+# from this repo at the same tag (build.sh --branch also accepts a tag ref).
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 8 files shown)


kilo-gi:
I'm going to click merge on this because 
1. it's basically a script and independent of the rest of the repo
2. github actions have to be on main in order to be able to run+test, so I have to merge to test end to end

PR #335, opened 2026-09-01
DNS refactor

zplizzi:
cleanup DNS. make it have one interface and not lots of entanglements. and make its state be adjustable at runtime, vs static based on config.


compute_space/src/compute_space/core/dns/coredns_provider/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""The router's own DNS: CoreDNS, the zones it answers for, and the records it serves.
```
zplizzi: simplify this to a single directive: use only the interface, and this package should not import from the rest of the code.

compute_space/src/compute_space/core/ip.py
```diff
@@ -0,0 +1,90 @@
...
+def is_public_ipv4(candidate: str) -> bool:
+    """``is_global`` consults the IANA special-purpose registry, so RFC 1918, loopback,
+    link-local, CGNAT, and the documentation ranges are all rejected in one check."""
+    try:
+        return ipaddress.IPv4Address(candidate.strip()).is_global
+    except ValueError:
+        return False
+
+
+async def detect_public_ip(
```
zplizzi, line 34: btw this isn't currently used, but i had it from a previous PR and it seemed nice so i figured i'd include.

zplizzi:
TODO: check if any docs should be updated based on these changes

zplizzi:
tested on my instance and seems to work, altho presumably didn't need to exercise the cert logic.

compute_space/src/compute_space/core/dns/coredns_provider/coredns.py
```diff
@@ -0,0 +1,188 @@
...
+from compute_space.core.dns.coredns_provider.records import DnsRecord
+from compute_space.core.logging import logger
+
+_TEMPLATES_DIR = Path(__file__).parent / "templates"
+# StrictUndefined so a template referencing a variable/attribute we forgot to pass raises instead
+# of silently rendering an empty string (e.g. a blank `file` path that CoreDNS would reject).
+_jinja_env = Environment(loader=FileSystemLoader(_TEMPLATES_DIR), undefined=StrictUndefined)
+
+# upstream resolvers for the container-facing DNS view's catch-all
+UPSTREAM_DNS = ("8.8.8.8", "1.1.1.1")
```
andrewlaack: Hard-coding public resolvers breaks private DNS and networks requiring their provided resolver. Could we retain host resolver discovery and use these as fallbacks?

compute_space/src/compute_space/core/dns/coredns_provider/interface.py
```diff
@@ -0,0 +1,211 @@
...
+            # these already resolve without coredns
+            return
+        if not self.serves_public_zones:
+            raise DnsNotEnabled(f"No address to serve {zone!r} on; this instance is not bound for public DNS")
+        normalized_zone = normalize_zone(zone)
+        if normalized_zone in self._zones:
+            # Already authoritative for it.  Appending anyway would put two server blocks for the
+            # same zone in the Corefile, which CoreDNS refuses to load.
+            return
+        await self._apply_zones((*self._zones, normalized_zone))
```
andrewlaack: Zone changes snapshot `_zones` before acquiring the lock. Three overlapping changes can drop a zone from CoreDNS while it remains in the DB. Could the read-modify-write happen inside the lock?

andrewlaack (approved):

PR #313, opened 2026-08-24
improve docs

zplizzi:

PR #336, opened 2026-09-01
Unify cross-app data access manifest

andrewlaack:
Uses access_all_app_data as the only public cross-app data permission


zplizzi (approved):
looks good but can you put it on top of https://github.com/cloud-in-a-bottle/cloud-in-a-bottle/pull/313 - some docs changes will conflict and i think would be easier to update here.

zplizzi:
(merged that PR)

PR #337, opened 2026-09-01
Allow changing the primary domain

andrewlaack:
Adds a Make primary action for configured domains and keeps certificate and DNS assets attached to the correct domains across changes. The operation atomically queues active apps, recreates them serially from deployed images with the new domain environment, leaves stopped apps stopped, and resumes safely after interruption. It coordinates app restarts with archive migration, storage enforcement, and concurrent app operations


andrewlaack:
this is too complex bc it tries to do stuff online with rollbacks and other bad decisions. Will  break this up and simplify.

PR #339, opened 2026-09-01
web: preserve request port in login redirects & app links (OH-275)

kilo-gi:
Follow-up from the VM-image work (#333). Rebased onto latest `main` (single commit).

## Problem

The distributed VM image is reached over a **wildcard-to-loopback domain** — `lvh.me` and `*.lvh.me` both resolve to `127.0.0.1`, so the dashboard is `lvh.me:PORT` and apps are `<app>.lvh.me:PORT`, all through one SSH-tunnel / NAT-forwarded port. That works with port **80** (no port to lose), but on any **non-default port** (`:8088` tunnel, `:8080` forward) the router builds absolute links/redirects from the bare domain name (`lvh.me`) and **drops the port** — so `/login` bounces the browser to `lvh.me:80` and dies. Reported hitting exactly this on the login redirect.

## Fix

Record the arriving origin once and carry its port onto every absolute URL the router builds back to this instance.

- **`RequestOrigin` (`web/helpers/zone.py`)** — a small value (arriving `Domain` + `host:port`) that `SubdomainProxyMiddleware` records per request in a `ContextVar`. It replaces the old `ZONE_SCOPE_KEY` request-scope stash. A `ContextVar` (not the request scope) because some link builders run *outside* the Jinja request context — notably `app_url()` called inside an *imported* macro (`app_row`, `nav_menu`), which Jinja renders without `request`, and which is exactly where the domain and port were silently going missing.
- **`host_with_request_port(host, request_netloc)` (`core/domains.py`)** — copies the request's port onto a router-built host. No-port (default 80/443) requests are unchanged. Lives in core (next to `Domain`) so core-layer link builders can use it without importing web; re-exported from `web/helpers/zone.py`.

Applied at every spot that builds an absolute URL to this instance:
- **`build_login_url`** (`web/auth/auth.py`) — the `/login` redirect carries the arriving port.
- **`app_url`** template global (`web/app.py`) — dashboard/app links carry the arriving port (`lvh.me:8088` → `foo.lvh.me:8088`), including links rendered inside imported macros.
- **`_resolve_edit_app`** (`web/routes/pages/apps.py`) — the "Edit this app" provider action stays on the browsing domain + port.
- **`approve_grant_url`** (`core/service_interface/headers.py`) — the permission-approval URL is built on the domain + port the owner is actually browsing (derived from the consumer app's `Origin` header, threaded in from `web/routes/services_v2.py`, since the router sees an internal-host service-proxy call here), falling back to the primary when there's no browsing context.
- **OAuth-return targets** (`web/routes/api/apps.py`, `add_app` / `reload_app`) — via `_oauth_return_host`, now that these run in the request coroutine and so can read the recorded origin.

No-port (default 80/443) requests are unchanged, so **TLS / multi-domain behavior is untouched** — the existing `build_login_url` tests still pass verbatim.

## Scope / not included

- **Apps still route by subdomain only** (there's no path-based app routing); this PR keeps that working over `*.lvh.me:PORT`, it doesn't add localhost access.

## Tests

- `test_multidomain_urls.py`: port-preservation cases for `build_login_url` plus a direct `host_with_request_port` table and `RequestOrigin` host builders.
- `test_multidomain_routing.py`: `zone_for_request` / origin-recording behavior.
- `test_nav_menu.py`: `app_url` carries the origin (domain + port) with no `request` in the Jinja context, and falls back to the primary when no origin is recorded.
- `test_services_v2.py`: `approve_grant_url` and `_oauth_return_host` carry the browsing domain + port and fall back to the primary without an origin.
- conftest resets the origin `ContextVar` around every test. ruff + mypy clean.

Boot-verified on a KVM host: image built from this branch, reached via `lvh.me:<port>`, `/login` redirect `Location` now carries the port.

### DIFF
--- a/compute_space/src/compute_space/core/service_interface/headers.py
+++ b/compute_space/src/compute_space/core/service_interface/headers.py
@@ -14,6 +14,8 @@ from urllib.parse import urlencode
 from compute_space.core.app_id import ROUTER_APP_ID
 from compute_space.core.app_id import ROUTER_APP_NAME
 from compute_space.core.auth.permissions_v2 import get_granted_permissions_v2
+from compute_space.core.domains import Domain
+from compute_space.core.domains import host_with_request_port
 from compute_space.core.domains import primary_domain_or_none
 from compute_space.core.service_interface.builtin_services import Permissions
 
@@ -64,16 +66,23 @@ def grants_for_provider(consumer_app_id: str, service_url: str, provider_app_id:
     ]
 
 
-def approve_grant_url(consumer_app_id: str, service_url: str, grant: Any, db: sqlite3.Connection) -> str:
-    """The owner-facing page for approving a grant a provider asked for."""
+def approve_grant_url(
+    consumer_app_id: str, service_url: str, grant: Any, db: sqlite3.Connection, browsing_netloc: str | None = None
+) -> str:
+    """The owner-facing page (on the router) for approving a grant a provider asked for.
+
+    Built on the domain the owner is actually browsing — ``browsing_netloc`` is the consumer
+    app's ``Origin`` authority on a browser-driven call — so it stays on their domain rather
+    than jumping to the canonical primary.  A server-side call has no Origin, so
+    ``browsing_netloc`` is None and we fall back to the primary.  Scheme follows the zone so a
+    plain-http (``.local``) instance builds a correct URL.
+    """
     # urlencode each value: service_url contains "/" and ":", grant is JSON with "{", "}", ","
     # and '"' — all of which break query-string parsing if interpolated raw.
     query = urlencode({"app": consumer_app_id, "service": service_url, "grant": json.dumps(grant, sort_keys=True)})
     approve_path = f"/approve-permissions-v2?{query}"
-    # Cross-app approval is server-side (no browsing request in hand), so this stays on the
-    # canonical/primary domain; use its scheme rather than a hardcoded https so a plain-http
-    # primary (e.g. a `.local` instance) builds a correct URL.
-    primary = primary_domain_or_none(db)
-    if primary is None:
+    zone = (Domain.match(db, browsing_netloc) if browsing_netloc else None) or primary_domain_or_none(db)
+    if zone is None:
         return approve_path
-    return f"{primary.scheme}://{primary.name}{approve_path}"
+    host = host_with_request_port(zone.name_no_port, browsing_netloc or "")
+    return f"{zone.scheme}://{host}{approve_path}"
--- a/compute_space/src/compute_space/core/service_interface/headers.py
+++ b/compute_space/src/compute_space/core/service_interface/headers.py
@@ -81,8 +81,14 @@ def approve_grant_url(
     # and '"' — all of which break query-string parsing if interpolated raw.
     query = urlencode({"app": consumer_app_id, "service": service_url, "grant": json.dumps(grant, sort_keys=True)})
     approve_path = f"/approve-permissions-v2?{query}"
-    zone = (Domain.match(db, browsing_netloc) if browsing_netloc else None) or primary_domain_or_none(db)
-    if zone is None:
+    matched = Domain.match(db, browsing_netloc) if browsing_netloc else None
+    if matched is not None:
+        # Browser-driven: keep the owner's domain and the access port they arrived on.
+        host = host_with_request_port(matched.name_no_port, browsing_netloc or "")
+        return f"{matched.scheme}://{host}{approve_path}"
+    # Server-side (or an origin we don't recognize): the canonical primary, its configured name
+    # verbatim so a port baked into the primary (e.g. `lvh.me:8080`) is preserved.
+    primary = primary_domain_or_none(db)
+    if primary is None:
         return approve_path
-    host = host_with_request_port(zone.name_no_port, browsing_netloc or "")
-    return f"{zone.scheme}://{host}{approve_path}"
+    return f"{primary.scheme}://{primary.name}{approve_path}"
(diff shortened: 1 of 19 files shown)


kilo-gi:
### Boot verification ✅

Built an image from this branch and booted it on a KVM host, claimed it, then hit `/` unauthenticated with various `Host` headers — the `/login` redirect now carries the request's port (and it's dynamic, not hardcoded):

| Request `Host` | `Location` |
|---|---|
| `lvh.me:8892` | `http://lvh.me:8892/login?next=http%3A%2F%2Flvh.me%3A8892%2F` |
| `lvh.me:9999` | `http://lvh.me:9999/login?next=…9999…` |
| `lvh.me` (no port) | `http://lvh.me/login?next=http%3A%2F%2Flvh.me%2F` (unchanged) |

The `next=` target keeps the port too, so the full login round-trip stays on the right port. Default-port (80/443) requests are byte-identical to before.

kilo-gi:
### Crawl verification (router + all default apps)

Built the image from this branch, claimed it, waited for all 6 default apps to reach *running*, then crawled every reachable page via `Host:` headers (so `*.lvh.me` subdomain routing works without DNS), extracting **every** absolute `lvh.me` URL from attributes, JSON blobs, and inline JS and flagging any not on the access port.

- **100 pages crawled** across the router pages **and** all six default apps (backup, catalog, filestash, oauth-provider, secrets, bottle-community-chat) — 0 app render errors.
- **Every real navigation link carries the port**, e.g. the dashboard app rows: `http://backup.lvh.me:8890/`, `http://catalog.lvh.me:8890/`, … and `app_detail` links.
- The only URLs missing the port are **literal example strings in the docs** (`docs/src/deploying.md`: `http://myapp.lvh.me:8080`, `http://lvh.me:8080/setup?claim=<token>`) — illustrative text, not links to the instance. Left as-is.

Covers the reported app-row regression and confirms the ContextVar approach reaches links rendered inside imported macros.

compute_space/src/compute_space/web/helpers/zone.py
```diff
@@ -1,23 +1,86 @@
...
+
+    def subdomain_host(self, label: str) -> str:
+        """An ``<label>.<zone>`` host carrying the request's access port."""
+        return host_with_request_port(f"{label}.{self.zone.name_no_port}", self.netloc)
+
+
+# The origin the current request arrived on, or None if none was recorded (no request
+# in flight, or a request that never matched a configured domain).  Set once per
+# request by SubdomainProxyMiddleware.
+_request_origin: ContextVar[RequestOrigin | None] = ContextVar("openhost_request_origin", default=None)
```
zplizzi: this (stashing in a contexvar) feels like not the right way to do this - can you look at this chat and see if we can do it in a more litestar-native way? https://claude.ai/share/eb9bb5b2-1de0-4c18-8243-c567789c0119. ie probably modifying/redoing _template_globals in app.py to use the engine_callback to get access to `context["request"]`. then i think you can just access the existing `scope[ZONE_SCOPE_KEY]` (or some variant of that) from `context["request"].scope["key"]` or something like that?

zplizzi (changes requested):

compute_space/src/compute_space/web/routes/api/apps.py
```diff
@@ -282,7 +294,7 @@ async def clone_and_get_app_info(
     if not repo_url:
         raise ValidationException(detail="No repository URL provided")
 
-    add_app_url = f"//{primary_domain(db).name}/add_app?repo={repo_url}"
+    add_app_url = f"//{_oauth_return_host(db)}/add_app?repo={repo_url}"
```
andrewlaack: Could these return URLs include the origin scheme too? GitHub's device flow redirects from an HTTPS page, so `//myhost.local:8088/...` resolves as HTTPS and breaks OAuth started from an HTTP secondary domain.

compute_space/src/compute_space/core/service_interface/headers.py
```diff
@@ -64,16 +66,23 @@ def grants_for_provider(consumer_app_id: str, service_url: str, provider_app_id:
...
-    # Cross-app approval is server-side (no browsing request in hand), so this stays on the
-    # canonical/primary domain; use its scheme rather than a hardcoded https so a plain-http
-    # primary (e.g. a `.local` instance) builds a correct URL.
-    primary = primary_domain_or_none(db)
-    if primary is None:
+    zone = (Domain.match(db, browsing_netloc) if browsing_netloc else None) or primary_domain_or_none(db)
+    if zone is None:
         return approve_path
-    return f"{primary.scheme}://{primary.name}{approve_path}"
+    host = host_with_request_port(zone.name_no_port, browsing_netloc or "")
```
andrewlaack: Could we preserve `zone.name` when `browsing_netloc` is absent? On the documented `lvh.me:8080` setup, server-side permission requests generate `http://lvh.me/approve-permissions-v2...`.

compute_space/src/compute_space/web/helpers/zone.py
kilo-gi: I had the same thought and had already talked with my agent about it - My understanding was that we had to put a contextvar somewhere to make it available to the Jinja template macros specifically, which don't have the request object passed in. And then given that we have the var, we might as well use the same var/function in the other places we do this?

I'm sure we can narrow this to just setting it around the jinja templates. I'll at least do that and I'll look at if there is a way to have the request object passed through to the macros. (or if we need macros specifically).

kilo-gi: ok yeah I don't know why it reached so hard for the contextvar - piping the context through was a lot simpler, good feedback.

I think this is resolved now.

zplizzi (approved):
lgtm at a glance - go ahead and merge and i'll give it a closer look later.

PR #343, opened 2026-09-02
e2e: read MinIO creds from its own container, not via file-browser

kilo-gi:
## What

`test_13e_read_minio_credentials` (and the whole archive chain `13f`–`13k` that cascades from it) went red on `main`. This makes the credential read namespace-independent and unblocks the chain.

`test_13e` read minio's `root-credentials.txt` through **file-browser**'s HTTP surface. minio writes that file `0600` (`umask 077`) as container-root; apps share podman's single rootless userns with `:idmap` bind mounts, so file-browser reading another app's `0600` file is a **cross-namespace read** whose success depends on idmapped-mount / kernel behaviour on the host.

This reads the file from **inside minio's own namespace** via `podman exec openhost-minio cat /data/app_data/minio/config/root-credentials.txt` over SSH — always container-root-readable, deterministic. minio is already confirmed running by `13c` and `start.sh` writes creds *before* launching the server, so no polling is needed. Also extracts the ec2 SSH invocation (already used by `13f`) into an `ssh_host` helper.

## Why it broke (trigger still unconfirmed)

This is **not** a code regression:
- last-green → first-red on `main` is a **docs-only diff**
- podman (**5.8.3**) and crun (**1.27**) are pinned in `pixi.lock` and **unchanged**
- upstream `minio/minio:latest` is ~a year old
- **both** gcp and ec2 (pinned AMI) fail identically

The exact environment trigger (kernel / idmapped-mount behaviour on the fresh instances) is **not yet identified** — we don't record kernel/podman versions in the e2e run. This change removes the fragile cross-app read regardless of the trigger.

> ⚠️ If file-browser genuinely can no longer read other apps' `app_data`, that's a separate product concern (that's file-browser's whole purpose) worth its own investigation. This PR only makes the *test* robust.

## Test

Not run live (needs a deployed cloud host). Verified: ruff clean, module collects (61 tests), no new mypy errors (`tool.mypy.files` excludes `tests/`).

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -786,9 +786,15 @@ class TestSelfHost:
         it is always container-root-readable) is namespace-independent and
         deterministic.
         """
+        # podman is installed inside the pixi env, not on the login PATH, and a
+        # non-interactive SSH shell never sources .bashrc — so a bare `podman`
+        # is "command not found".  Run it the way the openhost service does:
+        # `pixi run` from the code dir, which puts the pixi env's podman on PATH
+        # and loads its containers.conf / OCI-runtime config from the prefix.
         result = ssh_host(
+            "cd /home/host/openhost && /home/host/.pixi/bin/pixi run "
             "podman exec openhost-minio cat /data/app_data/minio/config/root-credentials.txt",
-            timeout=60,
+            timeout=120,
         )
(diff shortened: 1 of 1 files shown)


andrewlaack:
Noting this here: The real issue was a change I made to the filestash application which has since been resolved. Despite this, robustness improvements are warranted.

PR #357, opened 2026-09-03
Separate image builds from app launch

andrewlaack:
Extracts existing-image app launch from image building while preserving current behavior. This is the first foundation for restart-driven primary-domain promotion by making fresh container creation reusable without rebuilding.

### DIFF
--- a/compute_space/src/compute_space/core/apps.py
+++ b/compute_space/src/compute_space/core/apps.py
@@ -29,7 +29,7 @@ from compute_space.core.containers import stop_app_process
 from compute_space.core.containers import stop_container
 from compute_space.core.data import deprovision_data
 from compute_space.core.data import deprovision_temp_data
-from compute_space.core.data import provision_data
+from compute_space.core.data import make_data_dirs_and_env_vars
 from compute_space.core.data import rmtree_with_sudo_fallback
 from compute_space.core.domains import Domain
 from compute_space.core.domains import primary_domain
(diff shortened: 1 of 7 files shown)


compute_space/src/compute_space/core/apps.py
```diff
@@ -522,6 +488,103 @@ def _load_port_mappings_from_db(app_id: str, db: sqlite3.Connection) -> list[Por
...
+    app_row = db.execute("SELECT * FROM apps WHERE app_id = ?", (app_id,)).fetchone()
+    if app_row is None:
+        raise RuntimeError(f"No app found with id {app_id}")
+
+    app_name = app_row["name"]
+    local_port = app_row["local_port"]
+    container_id: str | None = None
+    launch_started = False
+    try:
+        env_vars = provision_data(
```
zplizzi: can we rename this to like `make_data_dirs_and_env_vars` or something? `provision_data` sounds like it might have meaningful side effects and thus would be weird to call twice on the `insert_and_deploy` path as is happening now.

compute_space/src/compute_space/core/apps.py
```diff
@@ -522,6 +488,103 @@ def _load_port_mappings_from_db(app_id: str, db: sqlite3.Connection) -> list[Por
     return [PortMapping(label=r["label"], container_port=r["container_port"], host_port=r["host_port"]) for r in rows]
 
 
+def launch_app_image(
```
zplizzi: nit: maybe `run` instead of `launch`? i think that's the standard verb for starting a container.

compute_space/src/compute_space/core/apps.py
zplizzi, line 687: nit: this is rather duplicative with deploy_app_background, i wonder if there's an easy way to collapse these together. if not easy, skip it.

zplizzi (approved):
lgtm!

PR #358, opened 2026-09-03
Reuse images for app restarts

andrewlaack:
Depends on #357.

Makes app restarts recreate containers from existing images with fresh runtime configuration, and makes startup resume durable restarts safely. This supplies the normal restart path that primary-domain promotion will use.


compute_space/src/compute_space/core/apps.py
```diff
@@ -691,17 +692,17 @@ def stop_running_archive_apps(
 
 
 def start_apps_by_id(app_ids: list[str], db: sqlite3.Connection, config: Config) -> None:
```
zplizzi: should this fn get renamed to `restart` also?

zplizzi (approved):

PR #359, opened 2026-09-03
Allow promoting the primary domain

andrewlaack:
Depends on #335 and #358.

Adds bare-minimum restart-driven primary-domain promotion, including stable certificate ownership, atomic app restart intent, safe planned restarts, and generation-aware Settings UI. This completes the initial primary-domain promotion effort without a live rolling swap.


andrewlaack:
Went through the full process, starting from an https domain -> an http only domain pointed at the instance (for testing purposes)

compute_space/src/compute_space/core/domains.py
```diff
@@ -189,6 +225,77 @@ def is_primary_domain(db: sqlite3.Connection, name: str) -> bool:
     return primary is not None and name.split(":")[0] == primary.name_no_port
 
 
+def domain_uses_legacy_cert_paths(db: sqlite3.Connection, name: str) -> bool:
```
zplizzi: i would prefer to do a system migration to put the certs into the new places vs scattering logic for dealing with the old paths all over the code? even if we mess that up it'd just automatically get a new cert at the next boot.

compute_space/src/compute_space/web/routes/api/archive_backend.py
```diff
@@ -330,16 +332,33 @@ def _quiesce() -> None:
...
-        await asyncio.to_thread(_run)
-    except RuntimeError as exc:
-        # 409 if it was a TOCTOU race against another configure attempt;
-        # 500 for genuine bring-up failures (detail is masked, so it rides in extra).
-        if "already configured" in str(exc):
-            raise ConflictException(detail=str(exc), extra={"code": "already_configured"}) from exc
-        raise InternalServerException(
-            detail="Failed to configure archive backend", extra={"output": str(exc)}
-        ) from exc
+        if db.execute(
```
zplizzi: would the archive_configuration lock not already cover this? this pattern feels weird to me - it's tough if every piece of the code needs to correctly check all the state / acquire locks from all the other parts of the code that it might interact with. and in this case it seems like we have two different lock-ish methods covering the same interaction.

compute_space/src/compute_space/web/routes/api/domains.py
```diff
@@ -216,6 +237,80 @@ async def add_domain(
...
+) -> Response[DomainListResponse]:
+    """Promote an existing ready domain, then restart through the standard startup path."""
+    if is_shutdown_pending():
+        raise ConflictException(detail="the instance is already restarting", extra={"code": "restart_pending"})
+    name = name.strip().lower()
+    record = get_record(db, name)
+    if record is None:
+        raise NotFoundException(detail="domain not found")
+    current_primary = primary_domain(db).name_no_port
+    expected_primary = data.expected_primary.split(":")[0].lower()
```
zplizzi: nit: why do we need to pass in / check expected_primary?

compute_space/src/compute_space/core/operation_locks.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import annotations
+
+import threading
+
+archive_configuration = threading.Lock()
```
zplizzi: general comment: i'm trying to switch to using async vs threading when possible, for consistency. which would mean async locks vs threading locks. if not easy here, that's fine.

zplizzi: altho i'm not actually sure how async event loops and threaded code will interact on a threading.Lock. your set_primary_domain appears to run from the normal async event loop

compute_space/src/compute_space/core/updates.py
zplizzi, line 104: for the future: we should probs request a restart from system updater, so it can hard kill us if we don't manage to shutdown. python is very bad at cleanly initiating its own shutdown.

zplizzi (approved):

PR #361, opened 2026-09-04
Add deploy redirect landing page

andrewlaack:
Adds an authenticated, side-effect-free /redirect/deploy landing page that prefills a repository without cloning it until the owner clicks Deploy. Existing /add_app prefill behavior remains unchanged. This enables portable deploy links through my.cloudinabottle.org.

Validation: targeted pytest (4 passed); Ruff check and format; git diff --check; two vet passes with no issues.


andrewlaack:
This is just an idea, hasn't been reviewed / discussed.

PR #369, opened 2026-09-05
Only point containers at the gateway resolver when it exists

zplizzi:
## The bug

`run_container` passes `--dns 10.200.0.1` to every non-`network_host` app. That
is where the container-facing CoreDNS view binds **on a real instance**, once
ansible has created the `openhost0` dummy interface. Nowhere else has that
address: not in dev, not in CI, and not on macOS, where the app test harness
runs the router on the host and containers inside the podman VM.

A container pointed at a resolver that isn't there has **no DNS at all**:

```
podman run --rm --dns 10.200.0.1 python:3.12-slim getent hosts pypi.org   # fails
podman run --rm                  python:3.12-slim getent hosts pypi.org   # resolves
```

Anything an app does over the internet breaks. Two we hit while testing
`imbue-hosted-spaces` locally:

- a Stripe call from the app container returned 500 with the log ending at the
  outbound request — the name lookup never completed
- vm-manager's entrypoint runs `uv run`, which reaches pypi, so the container
  never came up. The router reports this as `App started but not responding to
  HTTP`, which points nowhere near DNS.

It also silently breaks DNS for every app container on a developer's local
stack, not just test runs.

## The fix

Gate the flag on `is_bindable(CONTAINER_GATEWAY_IP)` — the same probe
`start.py`'s `_hairpin_gateway_ip` already uses to decide whether to *serve*
that view. The two sides now agree: we only point containers at the gateway
resolver when we're actually running one there. On a real instance nothing
changes; everywhere else containers keep podman's default resolver.

`start.py` already had the precedent, with the matching comment: *"Only add the
gateway bind if the interface actually exists (it won't in dev mode or CI where
openhost0 hasn't been created by ansible)."*

## Tests

`test_run_container_points_dns_at_gateway_for_pasta_apps` pins the probe True
(it would otherwise fail on any machine without `openhost0` — including CI), and
a new `test_run_container_skips_dns_when_gateway_is_absent` covers the other
branch. 90 passing across `test_containers.py` + `test_coredns.py`.

### DIFF
--- a/compute_space/src/compute_space/core/containers.py
+++ b/compute_space/src/compute_space/core/containers.py
@@ -29,6 +29,7 @@ from typing import NoReturn
 
 import attr
 
+from compute_space.core.ip import is_bindable
 from compute_space.core.logging import logger
 from compute_space.core.manifest import AppManifest
 from compute_space.core.manifest import PortMapping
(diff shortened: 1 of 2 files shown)


tucif:
Thanks for working on this! I ran into what looks like a related case that the current check might miss.

My instance is running in local HTTP mode (`lvh.me:8080`), with the intention of exposing it through Cloudflare Tunnel. After claiming the instance successfully, opening the app catalog failed with:

```text
fetch failed: Get "https://raw.githubusercontent.com/cloud-in-a-bottle/app-manifest/main/catalog.json":
dial tcp: lookup raw.githubusercontent.com on 10.200.0.1:53: i/o timeout
```

What I found on the host:

- `openhost0` exists and has `10.200.0.1/32`.
- `sudo ss -luntp 'sport = :53'` shows no DNS listener.
- `getent ahostsv4 raw.githubusercontent.com` resolves successfully.
- `dig @10.200.0.1 raw.githubusercontent.com +time=2 +tries=1` returns `connection refused`.
- The catalog container was started with `--dns 10.200.0.1`.

As a temporary workaround, I changed only the catalog container’s resolver to `1.1.1.1`/`1.0.0.1`, which restored catalog access.

I haven’t tested this PR itself, but reading the diff, it looks like `is_bindable(CONTAINER_GATEWAY_IP)` would still return true on my instance: the interface exists, but CoreDNS isn’t running.

Looking at the startup path, my suspicion is that `add_zone()` skips local-only domains such as `lvh.me`, so the internal resolver never starts, even though containers are configured to use it. Would it make sense to start the internal forwarding resolver independently of public zone serving, or fall back to Podman’s default DNS when that resolver isn’t enabled?

A regression test with **the gateway interface present, a local-only domain, and public DNS disabled** might help cover this variation. Happy to provide more details or test a fix!

zplizzi:
@andrewlaack does this fix the issue mentioned by @tucif ?

andrewlaack:
@zplizzi That’s a separate case where the gateway exists but CoreDNS is not running. It should be handled by starting the container forwarder independently of public DNS and authoritative zones in #372.

PR #380, opened 2026-09-14
Show a waiting page while apps start

andrewlaack:
Keep app links clickable and show a styled waiting page at the app URL during building and startup. Automatically retry safe page navigation until the app is ready, preserving deep links and existing authentication and failure behavior. The waiting page has no buttons or navigation controls.

Validated with focused proxy and browser/accessibility tests, pre-commit checks, and vet. Public-HTTPS live testing covered dashboard and Details launches, real startup and reload, mobile touch, one-time links, API requests, and failure handling. The latest control removal is covered by 103 proxy tests and both desktop/mobile browser cases.

Screenshot 2026-09-14 at 1 15 13 PM


compute_space/src/compute_space/web/helpers/app_starting.py
```diff
@@ -0,0 +1,74 @@
...
+from compute_space.web.helpers.static import make_static_url
+
+RETRY_SECONDS = 3
+_WEB_DIR = Path(__file__).resolve().parent.parent
+_JINJA_ENV = Environment(loader=FileSystemLoader(str(_WEB_DIR / "templates")), autoescape=True)
+_STATIC_URL = make_static_url(_WEB_DIR / "static")
+_QUALITY_VALUE = re.compile(r"(?:0(?:\.[0-9]{0,3})?|1(?:\.0{0,3})?)")
+
+
+def _is_html_navigation(request: Request[Any, Any, Any]) -> bool:
```
zplizzi: this fn, and file more generally, seems more complicated than i would have thought we need here. can we do something simpler?

compute_space/src/compute_space/web/helpers/app_starting.py
```diff
@@ -0,0 +1,74 @@
...
+            body=b"" if request.method == "HEAD" else b"Your app is coming up. Please try again shortly.",
+            status_code=503,
+            media_type="text/plain",
+            headers=headers,
+        )
+
+    router_host = host_with_request_port(zone.name_no_port, request.url.netloc)
+    router_url = f"{zone.scheme}://{router_host}"
+
+    def static_url(filename: str) -> str:
```
zplizzi: why do we need this here? seems like there ought to be some fn like this available to all jinja templates

zplizzi (approved):

PR #382, opened 2026-09-14
Fix e2e MinIO bucket creation after dl.min.io went 410 Gone

kilo-gi:
## What broke

`test_13f_create_minio_bucket` downloads the MinIO client at test time from `dl.min.io`. MinIO has archived the open-source Server, Client (mc) and KES projects and stopped serving those binaries, so that URL now returns **410 Gone** for every request and every architecture.

`curl -sL ... -o /tmp/mc` writes the 410 response *body* to the file and still exits 0, so `chmod +x` made the error page executable and the shell tried to run prose as a script:

```
/tmp/mc: line 1: 410: command not found
/tmp/mc: line 3: syntax error near unexpected token `('
```

Only `13f` actually broke. `13g`, `13h` and `13k` are cascades: `13f` never sets `_minio_bucket` / `_minio_endpoint`, so their `all([...])` guards go False and the archive backend still reports `local` instead of `s3`.

## The fix

Point `mc` at the `minio/mc` GitHub release assets, pinned to `RELEASE.2025-08-13T08-35-41Z`. The repo is archived but its release assets still serve real static binaries for both `linux-amd64` and `linux-arm64`, and pinning means no further upstream churn can move this.

Add `-f` so the next dead endpoint fails at the download with a readable curl error instead of a syntax error several commands later.

## This is not branch-specific

`main` is affected too. It only *looks* green because `e2e` runs on `push` and nothing has landed since 2026-09-10, which predates the takedown. The next push to main would have hit this.

## Verification

- Both pinned URLs return ELF binaries (`linux-amd64` 30.5 MB, `linux-arm64` 28.7 MB).
- The generated shell command expands correctly, arch detection included.
- Against the dead URL, old form exits 0 and writes ASCII text; new `-f` form exits 56 and writes nothing.
- Confirmed the 410 is deterministic, not a flake, so a CI retry could never have passed.
- The assertion path itself needs the cloud e2e job, which this PR exists to run.

## Follow-up worth considering

`mc` is used in exactly one place, only to create two buckets. That could be done with a single S3 `PUT /<bucket>` via `curl --aws-sigv4` (the VM is Ubuntu noble, so curl is new enough), removing the runtime third-party binary download entirely. Left out here to keep the unblocking fix minimal and reviewable.


kilo-gi:
Superseded by #383, which landed the better fix: it runs `mc` from inside the MinIO image instead of repointing the download, so nothing is fetched from an unversioned URL at all. Closing this in favour of that.

PR #389, opened 2026-09-15
diagnostics: disable the action bar until the snapshot loads

zplizzi:
## What

On the Diagnostics page, **Copy to clipboard** was clickable from first paint, but its handler starts with `if (!latest) return;` so an early click did nothing at all: no copy, no message, no sign the button had registered. **Refresh** and **Download JSON** had the mirror-image problem: clicking either while a collection was already in flight kicked off a second one (podman calls plus outbound reachability probes) on top of the running one.

All three now run off one `setActionsBusy()` in `diagnostics.js`:

- Copy and Refresh ship `disabled` in the template and are enabled by the script.
- While a fetch is in flight, all three are out of action.
- When it settles, Refresh and Download come back either way, so a failed load can be retried; Copy comes back only if there is a snapshot to copy (so it stays disabled after a failure).
- `latest` is cleared at the start of a load, since a reload means the page no longer holds a bundle matching what it is about to show.

The Download control is an `<a>`, which has no `disabled` attribute, so it uses `aria-disabled` (already styled by `.btn[aria-disabled="true"]` in `components.css`) plus a click guard that prevents the navigation. It is deliberately left enabled in the markup rather than disabled there: it is a plain server request and keeps working even if the script never loads.

## Testing

- New `compute_space/src/compute_space/tests/test_diagnostics_page.py` renders the page and asserts Copy and Refresh ship disabled while Download does not. Confirmed it fails if the `disabled` attribute is dropped.
- `pytest compute_space/src/compute_space/tests/test_diagnostics.py` and the new file: 95 passed.
- The repo has no JS test harness, so the script half was exercised by running the real `diagnostics.js` against a stub DOM in node (throwaway, not committed): disabled during first load, enabled after success, disabled again during refresh, Copy still disabled but Refresh usable after a failed refresh, all three enabled after a successful retry, and the Download click refused only while busy.
- A browser check was not possible in this container: `playwright install chromium` reports `Playwright does not support chromium on ubuntu26.04-x64` and no system chromium is available. The existing `tests/test_accessibility.py` axe scan of `/diagnostics/` runs after the page finishes loading, when every button is enabled again, so it should be unaffected; CI will confirm.

Linear issue: https://linear.app/imbue/issue/CB-375/diagnostics-page-copy-to-clipboard-should-be-disabled-till-the-page
Workspace: https://claude-workbench.host.zackpolizzi.com/?workspace=cloud-in-a-bottle/CB-375


zplizzi:
Browser check for this branch, filling the gap noted under Testing: chromium does run in the workbench container after `apt-get install` of the missing shared libraries (libglib2.0-0, libnss3, libgbm1 and friends); only `playwright install` refuses on ubuntu26.04-x64, and the already-cached browser works once those libs are present.

Drove `/diagnostics/` on `ae8f1f91` with playwright against a local stack, reading the live button state out of the DOM:

| | Copy | Refresh | Download `aria-disabled` |
|---|---|---|---|
| fetch in flight (delayed 3s) | disabled | disabled | `true` |
| after it lands | enabled | enabled | `false` |
| after an aborted fetch | disabled | enabled | `false` |

Clicking Download while `aria-disabled="true"` does not navigate: the page stays on `/diagnostics/`. `tests/test_accessibility.py` also passes locally on this branch (matching CI).

Two optional nits, neither a defect:

- While busy, the Download link keeps `tabindex=0` semantics, so it stays in the tab order and can be focused even though activating it is refused. Adding `tabindex="-1"` alongside `aria-disabled="true"` (and removing it after) would take it out of the tab order for the second or two it is out of action.
- `loadDiagnostics()` has no reentrancy guard of its own; it relies on Refresh being disabled to prevent overlap. That holds today, and an early `if (busy) return;` would make it hold for any future caller.

